PR #7, opened 2017-06-08
Rely on PATH for linuxsampler location

steveb:
This future-proofs the switch to linuxsampler-2 by finding the binary
whether it is installed in /usr/bin or /usr/local/bin from a source
install.


jofemodo:
Hi @steveb!
Zynthian is started as a service by systemd, and currently, no PATH variable is set, so your change will not work. 

I've added this line to zynthian.service unit, in the zynthian-sys repository:

`Environment=PATH=/bin:/usr/bin:/usr/local/bin`

In the next realease, i will push this change and yours.

PR #10, opened 2017-07-31
LS Engine upgrade, FS Engine optimizations and bugfixes

schpion:
The LS Engine have been tested with LinuxSampler 2.0.0 that was configured at compile time with following options:
`./configure --enable-max-voices=21 --enable-max-streams=64 --enable-stream-min-refill=4096 --enable-refill-streams=2 --enable-stream-max-refill=131072 --enable-stream-size=262144 --disable-asm --enable-subfragment-size=64 --enable-eg-min-release-time=0.001 --enable-eg-bottom=0.0025 --enable-max-pitch=2 --enable-preload-samples=65536`

The FS Engine now works well with big sound fonts.


jofemodo:
Great work, @schpion! Thanks a lot!

PR #19, opened 2017-11-04
Update zynthian_gui_admin.py

vbretsch:
I am running Zynthian in a german localized environment and ifconfig displays inet address in localized (german) form 'inet Adresse' which leads to the awk regexp not matching.
Therefore i propose to make the change above. Taken from here: https://stackoverflow.com/questions/6829605/putting-ip-address-into-bash-variable-is-there-a-better-way/14910952#14910952


jofemodo:
Thanks for the improvement!! ;-)

PR #45, opened 2018-07-08
Pianoteq engine improvements

Jonnie-Cache:
Added support for version 6.2 of Pianoteq and to permit the GUI to show more and better information about the version of Pianoteq that is installed.


jofemodo:
What version of Pianoteq are you using? It would be nice to detect "stage", " standard" or "pro".

Jonnie-Cache:
@jofemodo I don't know if the --version output differentiates between standard and pro. I only have the pro version so I have added the code to differentiate between stage and pro. If anyone can post the output of --version for the standard version, I may be able to improve the code further.

jofemodo:
OK! I've tested your changes and it work, so i merge with master.
If needed, we will add "standard" support when somebody with "standard" version can help with it.

PR #51, opened 2018-09-28
Pianoteq gui tweaks for Stage / Pro

josephfh:
Note: You will have to delete your preset cache to see the changes
`rm /zynthian/zynthian-my-data/pianoteq6/presets_cache.json`

Why?

* `<default>` presets not showing where preset has no name
* The bank list doesn't reflect the currently available instruments
* Demo and licensed instruments mixed together in the Pianoteq bank list

How?

* Fix the `<default>` presets by reworking the checking if statement
* Add missing instruments from the free KIViR.ptq pack
* Look for licensed instruments mentioned in the `<VALUE name="subl" val=""/>` line of the config XML
* Reorder the banks to _licensed_, then _free_, then _demo_
* Add a _spacer_ item to the bank list to separate full instruments from demo instruments

**To do:**

* I don't know what is used in the `<VALUE name="subl" val=""/>` line of the config XML for every instrument that can be purchased. I've only added them for the ones I can confirm. 
* First time coding in Python before so refactoring might be necessary

Tested with Pianoteq v6.3.0 STAGE

### DIFF
--- a/zyngine/zynthian_engine_pianoteq.py
+++ b/zyngine/zynthian_engine_pianoteq.py
@@ -80,6 +80,16 @@ def get_pianoteq_binary_info():
 		return res
 
 
+def get_pianoteq_subl():
+	subl=[]
+	if os.path.isfile(PIANOTEQ_CONFIG_FILE):
+		root = ET.parse(PIANOTEQ_CONFIG_FILE)
+		for xml_value in root.iter("VALUE"):
+			if(xml_value.attrib['name']=='subl'):
+				subl=xml_value.attrib['val'].split(';')
+	return subl
+
+
 def fix_pianoteq_config():
 	if os.path.isfile(PIANOTEQ_CONFIG_FILE):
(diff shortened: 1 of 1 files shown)


zyngine/zynthian_engine_pianoteq.py
```diff
@@ -231,6 +242,9 @@ class zynthian_engine_pianoteq(zynthian_engine):
 		('',0,'-- Other Instruments (DEMO) --')
 	]
 
+	if PIANOTEQ_VERSION[0]>=6 and PIANOTEQ_VERSION[1]>=3:
```
josephfh: Ant. petrof didn't exist before v6.3.0

jofemodo:
Nice work, Joseph! Thanks a lot for your contribution!
I would like to make "not clickable" the "== Other Instruments (DEMO) ==" line, but i will do after merging your changes ;-)

PR #69, opened 2019-03-24
Map presets to pedalboard presets

steveb:
Currently the zynthian presets are mapped to mod-ui effect presets, however
this mapping doesn't work very naturally in practice. It generally ends up with
a preset list where every preset is called 'Default' and it is not clear that
each preset might be for a different effect.

Fortunately mod-ui also has the concept of a pedalboard preset (possible this
didn't exist when the engine was first written). A pedalboard preset maps very
well to a zynthian preset, so this change replaces the effect preset with a
pedalboard preset implementation.

### DIFF
--- a/zyngine/zynthian_engine_modui.py
+++ b/zyngine/zynthian_engine_modui.py
@@ -67,6 +67,7 @@ class zynthian_engine_modui(zynthian_engine):
 		self.websocket = None
 		self.ws_thread = None
 		self.ws_preset_loaded = False
+		self.ws_bundle_loaded = False
 
 		self.bank_dirs = [
 			('_', self.my_data_dir + "/mod-pedalboards")
@@ -82,6 +83,7 @@ class zynthian_engine_modui(zynthian_engine):
 		self.graph = {}
 		self.plugin_info = OrderedDict()
 		self.plugin_zctrls = OrderedDict()
+		self.pedelpresets = OrderedDict()
 
 
 	def start(self):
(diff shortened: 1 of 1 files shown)


jofemodo:
Nice ... although i would like to also have the possibility of selecting "effect" presets. Perhaps we can combine both functionalities, adding the effect name to the preset name for better UX?

I mean ...

For MOD-UI engine, the list of "zynthian presets" would be composed of:

+ MOD-UI pedalboard presets when available
+ Separator "------------------"
+ Effect presets with format: #Effect Name#:#Preset Name#

In such a way, every use case is covered and everybody will be happy ;-)

What do you think? ;-)

steveb:
Yes, effect presets after the pedal presets would be nice, I'll take a look.

This raises a related issue, right now if a zynthian snapshot is selected, the mod-ui pedalboard is cleared and rebuilt, even if the only difference is a preset within the current pedalboard. This results in a noticeable cut in the audio. This isn't ideal for a guitar pedal to change presets mid-song.

Could zynthian compare current and next snapshots, and if they're identical except for the preset, just apply the preset?

jofemodo:
Yes, it could be done ... although it's a nice work ;-)
Also, it's not a good practice to save snapshots like that. It's a lot better to use ZS3 (subsnapshots), my friend. They are precisely thought for this use-case ;-)
Anyway, i take note of your suggestion and will try to implement a smarter loading procedure in the next code iteration ...

Regards,

jofemodo:
Hi @stevb!
Did you advance with this? I rellay would like to merge your changes, but plugin presets should be available too.

steveb:
I've got something locally which I'll clean up and push this weekend. Thanks for pointing out ZS3, that had escaped my attention :)

jofemodo:
Hi @steveb !

I forgot to told you that your changes work like a charm and your PR is merged with master now.
I'm really happy with this new feature!! Please, could you open a topic in the forum for announcing it and explaining a little bit how it works?

Thanks a lot!!

steveb:
> Please, could you open a topic in the forum for announcing it and explaining a little bit how it works?

sure thing :)

PR #75, opened 2019-04-10
I2C HWC

riban-bw:
Adds support for riban I2C HWC for rotary encoders.


riban-bw:
Wondering why this PR wasn't merged with the others relating to I2C HWC? Is it because this just sets defaults which will be set via configuration anyway, i.e. is not required? If so, should other similar defaults be removed?

jofemodo:
In fact is not required never more because webconf configures everything. I should delete/refactorize this code, but not today ... so i will merge your PR ;-)

PR #84, opened 2019-04-12
Feature/status font

### LINKED ISSUE #82: Use symbols in status bar
I would quite like to see symbols used in place of letters to indicate various status. Also use different colours to allow easier identification of message type (especially on smaller screens).

Use|Current | New | Unicode | Colour
----|---------|------|-----------|--------
Play | > | ► | 25BA | Green
Record | R | ● | 25CF | Dark Red
MIDI | M | ♫ | 266B | Blue
Over Temp | T | ▲° | 25B2, 00B0 | Bright Red
Under Voltage| V | ▼≈ | 25BC, 2248 | Bright Red

Would need to ensure each character renders appropriately in all available fonts. I don't know whether there is sufficient space to have two characters (as suggested for temperature and voltage).

It may be advantageous to change the border / background colour of a section of the screen (whole screen border or status area only) when an error occurs, e.g. X-run, over temp, under voltage, etc. This would differentiate from other fleeting but acceptable signals such as MIDI traffic or when a playback or recording starts or ends it can be difficult to differentiate the status change from coincident X-runs.

I can work up a PR for this if it is accepted as a desirable requirement.

riban-bw:
Fixes #82 - use icons / symbols in status bar.


jofemodo:
Nice! it's a good starting point for having a more "graphical" UI ;-)

Thanks!

PR #95, opened 2019-04-22
Feature/cuia recording

riban-bw:
Adds MIDI and audio record and stop functions to CUIA.

### DIFF
--- a/zynthian_gui.py
+++ b/zynthian_gui.py
@@ -477,19 +477,19 @@ class zynthian_gui:
 			
 		elif cuia == "START_AUDIO_RECORD":
 			if len(params) > 0 and params[0] > 0:
-				self.start_audio_record()
+				self.screens['audio_recorder'].start_recording()
 				
 		elif cuia == "STOP_AUDIO_RECORD":
 			if len(params) > 0 and params[0] > 0:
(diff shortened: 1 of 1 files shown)


riban-bw:
May I suggest that you use the PR mechanism within GitHub to discuss submissions, e.g. query why things are coded in a particular way, recommend or request changes like coding styles, etc? Also when merging it may be advantageous to squash the PR to avoid muddying the version history of the main branch. (There are often irrelevant commits during development of a patch.)

jofemodo:
Hi @riban-bw!

I suggest to avoid the creation of the new start/stop functions in the "zynthian_gui" class. 
The logging line is redundant:

`logging.info("Start audio recorder")`

because the start_recording/stop_recording functions already has a logging line. So better call directly from the "callable_ui_action" function to:

`self.screens['audio_recorder'].start_recording()`

`self.screens['audio_recorder'].stop_recording()`

Thanks!

jofemodo:
OK! I will use the "squash merge" ... 

Thanks for the advice! ;-)

riban-bw:
Good call. I added the extra functions before I found where / how to call the screen functions and added the logging to trace flow whilst debugging. Both removed now and much neater.

PR #96, opened 2019-04-23
Feature/toggle loop play

riban-bw:
Adds feature to toggle loop mode in audio and MIDI player.

### DIFF
--- a/zyngui/zynthian_gui_audio_recorder.py
+++ b/zyngui/zynthian_gui_audio_recorder.py
@@ -26,6 +26,7 @@
 import os
 import sys
 import logging
+import threading
 from time import sleep
 from os.path import isfile, isdir, join, basename
 from subprocess import check_output, Popen, PIPE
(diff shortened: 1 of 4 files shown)


jofemodo:
Hi @riban-bw !

It seems really good. Congratulations .... and thanks a lot!! ;-)

2 little things:

+ Coding Style: We're trying to improve coding style by "more or less" following PEP 8 (https://www.python.org/dev/peps/pep-0008/). Of course, there are lot of code to re-factorize yet, but we should try to do it with new code. For instance, please, leave 2 lines between function definitions ;-)

+ Try to not include changes from other PRs that aren't merged yet.. This PR include the changes from PR #95 ...

riban-bw:
Let me fix those issues and update the PR...

riban-bw:
I have made the requested changes. I interpret PEP 8 differently. It recommends:

> Surround top-level function and class definitions with two blank lines.

> Method definitions inside a class are surrounded by a single blank line.

So I would expect the `def` lines within each class to be separated by a single line.

Is there any plan to refactor all code to use spaces instead of tabs? Python 3 prohibits mixed space / tabs so it would need to be a concerted and holistic effort.

But for now, I hope this PR is acceptable. I will try to code to the current (most common) coding styles used in zynthian and as for guidance if I see deviation from PEP 8.

jofemodo:
You are right regarding the PEP 8 interpretation ... but i still think that 2 lines between "defs" inside classes improve readability and leaves more room for spacing code inside functions.

PEP 8 is a very good guide, but it also says that it should be adapted to the project ;-)

PR #97, opened 2019-04-27
Adds audio level monitoring

riban-bw:
Adds stereo peak audio monitoring.

- Status bar switchable between CPU / audio monitoring by setting ZYNTHIAN_UI_SHOW_CPU (True / False)
- Displays 2 horizontal bars indicating audio peak level
- Spans range -50dBFS to -10dBFS (green), -10dBFS to -3dBFS (yellow), -3dBFS to 0dBFS (red)
- Logarithmic range
- Damped decay

Not yet implemented:
- ~~Auto connect (must currently connect instruments using audio routing)~~ [Done]
- ~~Scale legend, e.g. vertical graduation marks~~ [Won't do]
- ~~Peak level hold, e.g. highest value showing as vertical line for a second~~ [Done]
- ~~No CPU / RAM bench marking yet performed~~ [Done]

Known issues:
- Low resolution screen (160 pixel wide) does not show red zone)

### DIFF
--- a/jackpeak/CMakeLists.txt
+++ b/jackpeak/CMakeLists.txt
@@ -1,0 +1,12 @@
+cmake_minimum_required(VERSION 2.6)
+project(jackpeak)
+
+include(CheckIncludeFiles)
+include(CheckLibraryExists)
+
+link_directories(/usr/local/lib)
+
+add_library(jackpeak SHARED jackpeak.h jackpeak.c)
+	target_link_libraries(jackpeak jack)
+
+install(TARGETS jackpeak LIBRARY DESTINATION lib)
--- a/jackpeak/LICENSE.txt
+++ b/jackpeak/LICENSE.txt
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.  We, the Free Software Foundation, use the
+GNU General Public License for most of our software; it applies also to
+any other work released this way by its authors.  You can apply it to
+your programs, too.
+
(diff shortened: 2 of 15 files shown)


riban-bw:
I have benchmarked against master and cannot detect increase in cpu usage when metering is added.
I can't figure out how autoconnect works so haven't been able to automatically feed the audio monitor. Please help!

riban-bw:
I tried adding graduation marks but it didn't look good. I added default routing of layer to level meters and peak hold indicators. The small issue with the red not showing on low resolution screens can be a follow-on action. I think this looks quite good now. Please review and feedback.

Areas I think might need attentions:
- Naming - jackpeak might not be the best name, maybe levelmeter or similar. The library does show peaks in a JACK input!
- There is a lot of code in the display. It looks clumsy to me but I am a C/C++ programmer so don't know whether Python is supposed to look like this!
- Is the range okay? I chose -50dB because it looks about right. Fluidsynth is rather noisy with noise floor sitting about -157dB (okay - not that noisy but it is worse than others). I don't think we need to see this noise on our meter (external monitoring could pick that up). 50dB below fsd seems a reasonable range though it could possibly be a little smaller.
- Add switch to toggle between CPU / audio level - maybe in webconf.

riban-bw:
Issue: Inserting an effect in a layer does not remove the routing of the engine from the meter. Need to figure this out (probably mod to autoroute?).

jofemodo:
Hi @riban-bw!

As seems to have a good git/github knowldge, i would like to comment with you about the best practices and workflow for improving zynthian development.

For instance, what is the best way of testing this pull request?

1. Clone locally and test your forked repo before merging with master
2. Merge your PR in a "testing" branch.
3. Other

Thanks a lot!

jofemodo:
Ahhh! I have invited you to be a "member" of zynthian team, so you can create branches and push to all but master ;-)

jofemodo:
I created a specific branch "gh93-levelmeter" and merge the pull request on it for testing.

jofemodo:
Hi @riban-bw !
It works really nice. Congratulations!!
A few adjustments for perfect integration:

+ Regarding the jack client name: jackpeak is OK for me. It's the original codename and clear enough for something quite techy that only developers needs to know ;-)

+ Regarding jackpeak itself: It seems quite simple and eficient. Anyway, perhaps some functions are not used nor needed: connect, disconnect & disconnectAll
I've tested without this functions and it works perfectly too. Less is more ;-)

+ Regarding the audio routing, it's better you do nothing in "zyngine/zynthian_layer.py" and implement the "magic" in the autoconnector. Please, remove this change and i will add jackpeak to the autoconnector.

jofemodo:
OK! Already commited some changes on the branch gh93-levelmeter ... 0;-)

PR #127, opened 2019-05-27
Add IntelliJ/Pycharm details into .gitignore

wyleu:
Just a niceness for us on PyCharm...


wyleu:
Move to VSC and away from Pycharm probably represents a developer drift.

Could be closed unless we configure .gitignore for vsc.

Issue #180, opened 2019-12-09
Audio Recorder plays MP3


mheidt:
Make the Audio Recorder display mp3 files that were copied into the captures directory and play them.

PR #181, opened 2019-12-09
#180 MP3 player

mheidt:
Easy addon that doesn't hurt anyone...


jofemodo:
Why not including ogg too?

mheidt:
Because I didn't needed it and I don't know what that player is capable of.

mheidt:
The only disadvantage I see is, that you see the converted ogg files as well, so that they are shown twice...as wav and ogg.
I can checkin the enhanced if statement.

jofemodo:
OK! Don't add ogg by now ...

PR #197, opened 2020-01-18
Sleep between calls to amixer

steveb:
My zynthian-ui was restarting without any useful logging whenever I loaded an
engine (tried mod-ui and linuxsampler).

I bisected the problem to commit 15eafb95cbaf680bf66aa1f0846da98a85bb6649,
shlex seems to be parsing the amixer commands just fine.

Logging the amixer output showed that some amixer commands didn't return an
output, but check_output also didn't raise any exception. Adding a sleep
between amixer calls seems to fix this issue for me.


steveb:
I think the overhead of shell=True was causing just enough delay to avoid this problem

jofemodo:
Thanks @steveb ! Good work!

PR #221, opened 2020-04-05
Add a radio player to zynthian

Baggypants:
Here is a terrible hack for how the radioplayer could work for issue zynthian/zynthian-issue-tracking#990 

I've tried to remove the record functionality, I don't know about the loop. It's possible someone could add a playlist tof local files and want to loop it I suppose.

The m3u8 or m3u file contents should look like
```
root@zynthian:/media/usb0/playlist# cat BBC6Music.m3u8
#EXTM3U
#EXTINF:-1, BBC 6Music
http://a.files.bbci.co.uk/media/live/manifesto/audio/simulcast/hls/uk/sbr_high/llnw/bbc_6music.m3u8
```

There could be possible integration in the webconf to add a web url and have it create the m3u8 file.

### DIFF
--- a/zyngui/__init__.py
+++ b/zyngui/__init__.py
@@ -22,6 +22,7 @@ __all__ = [
 	"zynthian_gui_confirm",
 	"zynthian_gui_main",
 	"zynthian_gui_audio_recorder",
+	"zynthian_gui_radio_player",
 	"zynthian_gui_midi_recorder",
 	"zynthian_gui_autoeq"
 #	"zynthian_gui_osc_browser"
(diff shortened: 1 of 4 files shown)


Baggypants:
Oh, and you'll need a way to create the playlist dirs otherwise things complain.

Baggypants:
I've found that it doesn't work if the playlist container another playlist e.g.

http://media-ice.musicradio.com/RadioXManchesterMP3.m3u

Even if I try `-allow-dangerous-playlist-parsing` Looks like I need to go back to the stations.csv that pyradio uses.

Baggypants:
I figured out a way to make it work, curl the `http://media-ice.musicradio.com/RadioXManchesterMP3.m3u` and put the resulting url in a `.pls`

PR #232, opened 2020-04-15
Radioplayer

Baggypants:
OK, There are two things here, one is a player to play internet radio, or any playlist, using the engine system so it can be passed thought the jack effects chain for amusement. It uses mplayer and is based on the audio recorder. There should be no extra dependencies.

The second is a front end for [rtl_fm_streamer](https://github.com/AlbrechtL/rtl_fm_streamer) which  isn't included in this pull request, it will have to be added as an install script. This enables users to use their rtl-sdr dongle with zynthian-ui as a fairly basic FM receiver and pipe it into the zynth plugin chain as above. It could be expanded to cover MW or amateur radio band support in the future.

This pull request is in response to issue zynthian/zynthian-issue-tracking#990 

I'd recommend it's pulled into it's own branch for testing to begin with as the sdr bit requires more stuff to happen.


jofemodo:
Hi @Baggypants!

I've been following your advances with this. Congratulations and thanks for your work. I will test it ASAP, although perhaps it takes some days. I'm quite busy and a there is a bunch of high priority tasks i have to attend ... but  will give it a try ASAP ... (i should get a  rtl-sdr dongle too! ;-) )

Baggypants:
Thanks dude! I'm totally fine with the of delay. I'm happy to make any changes you need.

If you do get an rtlsdr the bundled ariel will get you some stuff like aeroplanes, but you need better ariels for the amateur radio or dbtv stuff.

zyngine/zynthian_engine_radio.py
```diff
@@ -0,0 +1,292 @@
+# -*- coding: utf-8 -*-
+#******************************************************************************
+# ZYNTHIAN PROJECT: Zynthian Engine (zynthian_engine_puredata)
+#
+# zynthian_engine implementation for PureData
```
jofemodo, line 5: PureData Radio? ;-)

jofemodo:
Could you tell me what i need to install and configure for having everything working?
If you could make a PR with the recipes, it would be fantastic ...

riban-bw:
Assigning to @Baggypants to merge updates from testing branch and respond to feedback comments.

riban-bw:
@Baggypants, I assigned this back to you 4 months ago. I have just taken a more in depth look at the hardware requirements and it is a rather niche requirement that configures hardware in an unorthodox way, basically misusing it! I wonder whether there is significant benefit from adding support for this to the core application. Maybe it could be something that gets added as a community module when we implement that feature? Certainly the current PR does not seem to be moving at any real speed!

jofemodo:
I agree @riban-bw on implementing this as a community module, when this can be done. I don''t think this should be a core feature, although certainly it's an interesting and funny one. Currently Zynthian is quite overloaded of features, and we must be selective when adding new features to the core.

Anyway, FYI @Baggypants , we expect to do progress with the implementation of "community modules" in the next months.

Regards,

keinstein:
Is this issue still tested and/or related to the InternetRadio player current (2026/06, ORAM) Zynthian stable?

Maybe I'm mistaken, but since this issue is in „Testing“ stage in the Kanban board, I assume something is stuck. It seems not to be a “community module” since no reference to such a module is posted here after 5 years.

At least I interpret “testing” as: “It works, but we don't trust it, yet.”

BTW.: I came here, because I wanted to know where I can change the json file describing the internet radio stations.

PR #234, opened 2020-04-16
Revert "Stepseq jofemodo"

riban-bw:
Reverts zynthian/zynthian-ui#233. Does not work without associated jack midi clock.


riban-bw:
I merged this PR then realised there were issues so reverted but the PR remains closed. I am not sure how to change that... I have manually applied (most of) your stepseq changes but have refactored the code substantially so this will not merge. Will you please remove zynthian_gui_stepseq.py from the PR and resubmit so that I can merge just the other bits which are required for step sequencer?

PR #250, opened 2020-06-05
Show class for LV2 effects when selecting (+ more)

jypma:
This provides a convenient sorting and better discoverability for LV2 effects, especially if more than a few are enabled from the web UI.

For the moment, the class is displayed only as a prefix, but it might be better to create nice sub-headings for these, as is done on other screens.

It'd also be nice to eventually categorize the webconf this way as well.

---

_UPDATE_: This PR now also contains various other fixes / improvements I've been working on. See individual commit messages for details.

The commits themselves are quite small and can be reviewed individually, but please let me know if you'd like a different PR for some of them.

### DIFF
--- a/zyngine/zynthian_controller.py
+++ b/zyngine/zynthian_controller.py
@@ -373,6 +373,7 @@ class zynthian_controller:
 				self.set_midi_learn(int(snapshot['midi_learn_chan']), int(snapshot['midi_learn_cc']))
 		else:
 			self.set_value(snapshot,True)
+		self.refresh_gui()
 
 
 	#--------------------------------------------------------------------------
@@ -530,6 +531,9 @@ class zynthian_controller:
 	def midi_control_change(self, val):
 		value=self.value_min+val*self.value_range/127
 		self.set_value(value)
+		self.refresh_gui()
+
+	def refresh_gui(self):
 		#Refresh GUI controller in screen when needed ...
 		try:
 			if (self.engine.zyngui.active_screen=='control' or self.engine.zyngui.modal_screen=='control') and self.engine.zyngui.screens['control'].mode=='control':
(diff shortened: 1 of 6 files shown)


jofemodo:
I merged into "zynthian:jypma_lv2_display_class" branch for easy testing across the community ...

jofemodo:
Hi @jypma!

I'm testing your changes. First of all ... congratulations and thanks a lot!!

Let's go with the comments ;-)

**Regarding the first commit: adding the plugin "class" ...**

... I see that it's absolutely useful for Audio-FX and Generators, but i'm afraid it only adds noise for MIDI Synths & MIDI FX (this last engine-type is available on master branch from today!), so...

What would you think of applying this change only to Audio-FX & Generator engine types?
Also check if the class field is empty. We don't want names starting with " - " on the engine list.

jofemodo:
The 2 next commits seems OK to me:


+ Preselect current ZS3 program when opening menu efb6238
+ Update GUI when restoring a snapshot 4b4334b

Eyyy! You solve the refreshing problem when changing ZS3!!! ;-)

jofemodo:
Regarding this one:

+ Don't sleep when restoring ZS3 presets for LV2 6021984

I would limit to removing the sleep for LV2, but keep setting the bank and preset as it's needed for restoring the bank/preset names, those selected when the ZS3 was saved.

Could you measure how big is the improvement when avoiding only the sleep?

Thanks!

jypma:
Keeping the preset / bank changes is fine with me (they're empty on my presets since I don't work from presets). But I'm unsure about semantics / usability. If a user selects a preset, _then_ changes the controllers, it's no longer that preset... just not sure if it'd be correct to show it as such.

As to the sorting, you have some good points there. I can get a forum thread going. How do I create a screenshot from the zynthian screen? It'd be nice to include alternatives in the thread.

If an engine only has 1 class, we can do without it; but I was actually considering reworking the list to use non-selectable headers, instead of prefixes. A little like the "Waiting for program change" entry the ZS3 learn screen uses. That would also would work better for lower-resolution screens.

PR #251, opened 2020-06-10
Display LV2 class for configured plugin (and other small changes)

jypma:
This provides a convenient sorting and better discoverability for LV2
effects, especially if more than a few are enabled from the web UI.

Configured plugins are grouped by categories and then sorted. The
categories are displayed as non-selectable headers in a different
color.

The built-in synths are still displayed first in their original order.

*New Synth Layer*:
screenshot-synth
*New Audio-FX Layer*:
screenshot-fx
*New MIDI-FXLayer*:
screenshot-midi
*New Special Layer*:
screenshot-special2

---

This PR also contains a few other smaller fixes, see individual commits for details.

**TODO**
- [x] Handle missing CLASS on plugins more gracefully (even no included LV2 plugin is missing it)
- [x] (maybe) Do restore bank & preset on LV2 snapshots, since users may be expecting this

### DIFF
--- a/zyngine/zynthian_layer.py
+++ b/zyngine/zynthian_layer.py
@@ -523,7 +523,7 @@ class zynthian_layer:
 			}
 
 			for k in self.controllers_dict:
-				logging.info("Saving {}".format(k))
+				logging.debug("Saving {}".format(k))
 				zs3['controllers_dict'][k] = self.controllers_dict[k].get_snapshot()
 
 			self.zs3_list[i] = zs3
@@ -536,19 +536,18 @@ class zynthian_layer:
 		zs3 = self.zs3_list[i]
 
 		if zs3:
-			if zs3['bank_name']:
-				#Load bank list and set bank (if one was selected)
-				self.load_bank_list()
+			# Set bank and load preset list if needed
+			if zs3['bank_name'] and zs3['bank_name']!=self.bank_name:
 				self.set_bank_by_name(zs3['bank_name'])
+				self.load_preset_list()
 				self.wait_stop_loading()
 
-			if zs3['preset_name']:
-				#Load preset list and set preset (if one was selected)
-				self.load_preset_list()
+			# Set preset if needed
+			if zs3['preset_name'] and zs3['preset_name']!=self.preset_name:
 				self.set_preset_by_name(zs3['preset_name'])
 				self.wait_stop_loading()
 
-			#Refresh controller config
+			# Refresh controller config
 			if self.refresh_flag:
 				self.refresh_flag=False
 				self.refresh_controllers()
@@ -559,11 +558,11 @@ class zynthian_layer:
 			if not self.engine.nickname.startswith('JV'):
 				sleep(0.3)
 
-			#Set active screen
+			# Set active screen
 			if 'active_screen_index' in zs3:
 				self.active_screen_index=zs3['active_screen_index']
 
-			#Set controller values
+			# Set controller values
 			for k in zs3['controllers_dict']:
 				self.controllers_dict[k].restore_snapshot(zs3['controllers_dict'][k])
 
--- a/zyngine/zynthian_layer.py
+++ b/zyngine/zynthian_layer.py
@@ -523,6 +523,7 @@ class zynthian_layer:
 			}
 
 			for k in self.controllers_dict:
+				logging.info("Saving {}".format(k))
 				zs3['controllers_dict'][k] = self.controllers_dict[k].get_snapshot()
 
 			self.zs3_list[i] = zs3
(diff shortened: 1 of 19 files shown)


zyngine/zynthian_layer.py
```diff
@@ -535,27 +535,31 @@ def restore_zs3(self, i):
...
+				#Load preset list and set preset
+				self.load_preset_list()
+				self.set_preset_by_name(zs3['preset_name'])
+				self.wait_stop_loading()
+
+				#Refresh controller config
+				if self.refresh_flag:
+					self.refresh_flag=False
+					self.refresh_controllers()
+				sleep(0.3)
```
jofemodo, line 556: Please, keep setting the bank and preset. When dealing with synths, most of times you use a preset, and although you make some adjustments, you want to know you are using this preset when doing a ZS3 change. Only the sleep can be avoided.

zyngui/zynthian_gui_engine.py
```diff
@@ -43,26 +43,23 @@
...
-
-@initializator
 class zynthian_gui_engine(zynthian_gui_selector):
 
 	single_layer_engines = ["BF", "MD", "PT", "PD", "AE", "CS"]
 	check_channels_engines = ["AE"]
 
-	@classmethod
-	def init_engine_info(cls):
+	def init_engine_info(self):
```
jofemodo, line 51: My fried, this *MUST* be a class method. Webconf call this function for getting the engine list without creating an instance ...

zyngine/zynthian_layer.py
jofemodo: Well, probably we could also avoid the load_banks, and if the bank doesn't change, the set_bank & load_preset too. If the preset is also the same, then we could avoid the load preset too, but we must add the checks ...

jofemodo:
BTW, i like a lot the engine "class" subheaders ... really! Good idea ... ;-)

zyngui/zynthian_gui_engine.py
jypma: Indeed. Turns out it's not only webconf, just loading a preset also doesn't work this way; I overlooked that while testing. I'll keep the class-level structure and see if we can do smarter filtering on instance level.

zyngui/zynthian_gui_engine.py
```diff
@@ -91,38 +100,61 @@ def __init__(self):
 	def set_engine_type(self, etype):
 		self.engine_type = etype
 		self.midi_chan = None
-
+		self.init_engine_info()
```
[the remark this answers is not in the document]
jypma: It's related to my attempt to get rid of the static data, which I'll revert (see my other comment below)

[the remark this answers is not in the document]
jypma: However, `set_engine_type()` _does_ affect `engine_type`, which affects the filtering, which is done statically in the current code... so how would the list view get updated correctly after `set_engine_type()` is called, currently?

zyngui/zynthian_gui_engine.py
```diff
@@ -91,38 +100,61 @@ def __init__(self):
...
+			else:
+				items = sorted(infos.items(), key = lambda kv: kv[0])
+			for en, info in items:
+				self.fill_list_item(en, info)
 
-				ei=self.engine_info[en]
-				self.list_data.append((en,i,ei[1],ei[0]))
-				i=i+1
+		if (len(self.list_data) == 0):
+			self.list_data.append((None,len(self.list_data),"Please use webconf to enable plugins."))
```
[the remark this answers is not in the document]
jypma: Actually, I'll see if I can just show an actual URL here; early users may not recognize the term "Webconf" either.

zyngine/zynthian_layer.py
jypma: I've brought them back, but skip over them if no bank or preset has ever been selected in that snapshot (which reflects my usage at least :-).

zyngui/zynthian_gui_engine.py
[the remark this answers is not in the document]
jofemodo: Filtering is not done statically. "fill_list()" do the filtering using "engine_type" and it's not a class/static method. Only "init_engine_info()" is a class method, and it builds an unfiltered list.

zyngui/zynthian_gui_engine.py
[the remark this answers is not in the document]
jofemodo: New users shouldn't see this message, as pre-built images should have a pre-configured selection of engines for every type (i think MIDI tool is currently empty, although this will change very soon).

zyngine/zynthian_layer.py
jofemodo: This seems OK. I will test it ASAP.

jofemodo:
Hi @jypma !
I fixed some problems with the zynthian_gui_engine new code and improved a little bit the "styling". I hope engine list subheaders are a little nicer now ...

jofemodo:
Hi @jypma !

I tested a little bit more and everything seems fine. I just merged with master ...
Please, do extra checking (on master) and confirm everything is OK for you.

Thanks a lot!!

jypma:
Thanks a lot for your help! I won't have time for a test until at least tomorrow though.

jypma:
Quick test of master seems OK. Thanks for merging it!

As to the change in colors, I do prefer the original colors I selected for the headings, for two reasons:
- The dark background would give a lot more contrast, making them easier to read. This is especially important on LCD screens (which have lesser contrast to begin with), and embedded device with may have awkward viewing angles, further reducing contrast.
- Giving the headers a little left margin (by using spaces) makes it easier to visually distinguish them from normal entries. Prefixing them with ">" negates that.

However, since it's now merged, let's just leave it as-is; functionally, it is working fine.

Issue #195, opened 2020-01-15
Per-layer audio-channels configuration: Stereo, Right, Left


jofemodo:
Currently, all layers are stereo. It would be desirable to have mono layers (Right or Left), so we can have 2 separated chain effects, or 2 synth layers with separated audio outputs, etc.

jypma:
Some ideas for UX (assuming 4 input channels and 4 output channels, for maximum complexity). Not all variants would be needed right away, I assume :-)

When a selection can be made (more input channels than FX inputs, or more playback channels than FX output), display the following menu(s).

## When selecting a mono FX as root
```
[CREATING MONO FX]
--- Select input: ---
Input (1)
Input (2)
Input (1 + 2)
Input (3)
Input (4)
Input (3 + 4)
```
where `(1 + 2)` would downmix into mono, just like right now.
## When selecting a stereo FX as root
```
[CREATING STEREO FX]
--- Select input: ---
Input (1 + 2)
Input (3 + 4)
```
where `(1 + 2)` means a normal stereo connection of those inputs to the FX.
## Adding a stereo FX to a mono chain
Hook the mono output up to both stereo inputs (we may be doing this already)
## Adding a mono FX to a stereo chain
Connect both stereo outputs to the mono input (we may be doing this already)
## When selecting a stereo FX as root, or when changing a mono FX chain to stereo:
```
[CREATING STEREO FX]
--- Select output: ---
Output (1 + 2)
Output (3 + 4)
```
## When selecting a mono FX as root, or when changing a stereo FX chain to mono
```
[CREATING MONO FX]
--- Select output: ---
Output (1)
Output (2)
Output (1 + 2)
Output (3)
Output (4)
Output (3 + 4)
```
where `(1 + 2)` would output the mono output to both playback channels.
This last option is for the common use case of having a mono guitar go into headphones outputs, which are stereo.

jofemodo:
I think the output configuration can be simplified to:

+ Mono
+ Stereo

Anyway, zynthian currently does not support more than 2 audio channels, and i prefer to keep it simple by now. Zynthian is not a DAW .... yet! ;-)

So configuration options for "audio input" would be:

+ Mono R
+ Mono L
+ Mono R+L => and we avoid the need for a "mono" root-plugin when wanting to do this.
+ Stereo

Regards,

jypma:
I totally get that, we'll wait for the "export Ardour channel strip configuration to Zynthian" feature another time ;-) 

[...]

PR #253, opened 2020-06-14
WIP Configure input capture channels for a layer (#195)

jofemodo:


zynautoconnect/zynthian_autoconnect.py
```diff
@@ -513,20 +512,41 @@ def audio_autoconnect(force=False):
...
-			if len(rl_in)>0:
-				if len(rl_in)==1:
-					rl_in.append(rl_in[0])
+			# First, disconnect all capture ports from all inputs
+			for cap in system_capture:
+				for c in range(0, len(rl_in)):
+					try:
+						jclient.disconnect(cap, rl_in[c])
+					except:
+						pass
```
jofemodo, line 532: Ups! It can't be done like that. For avoiding "clicks" we have to disconnect the inputs that are not connected, but never disconnect a currently connected pair. If not, every time you add/change an effect, everything get disconnected and reconnected, generating lot of noises, especially if there is sound flowing thru the chain.

jofemodo:
I merged into development branch "jypma_input_config" ...

zynautoconnect/zynthian_autoconnect.py
jypma: I am aware, and was a little afraid of that would be needed :) That's why I called it WIP, and didn't merge it anywhere yet.

We can totally do it; however, on the hifiberry I'm testing with, there were no huge clicks while playing around.

jypma:
@jofemodo Can you please let me know if you continue to work on this feature (as you wrote a few days ago), or leave it up to me? Just want to make sure there's no double work :-)

jofemodo:
I will be tracking & testing your changes, and eventually, i will take over for final adjustments before merging, but i will warn you for avoiding double-work.

BTW, i don't see the use-case for having the possibility of crossing channels (sending 1 to 2 and 2 to 1). Your proposed solution is elegant and works fine, but IMHO, it's not needed at all. Of course, if you show me the use case, i would change my mind ;-)

Less is more, and Zynthian-UI is quite complex already, so my current trend is to not introduce any improvement without a clear use-case.

To be more explicit:

Why not using simple checks (a [X] instead of [1] and [2]) for selecting the audio input channels?
In such a way, we would support these use-cases:

+ Mono-input plugin connected to both capture inputs, that would be mixed to mono
+ Mono-input plugin connected to one capture input (1 or 2)
+ Stereo-input plugin connected to both capture inputs
+ Stereo-input plugin connected to one capture input (1 or 2) that would be duplicated

Your solutions also would support these use-cases:

+ Stereo-input plugin cross-connected to both capture inputs
+ Stereo-input plugin connected ONLY to one capture input, keeping unconnected the other one

But ... i can't see the utility of these cases ;-)

Regards,

riban-bw:
Swapping channels is a useful feature but could be implemented within a LV2 plugin.

jypma:
@jofemodo Can you explain the UI you have in mind a bit further? I'm not sure how to simplify things.

Even without allowing swapping of stereo channels, you'd still need, for every input channel, to configure to which plugin input it should go. Take for example one of the cases you describe: "Stereo input plugin connected to one capture input". You'd need, for that capture input (2 in the example below), to select that it go to both plugin inputs:
```
[ ] capture 1
[plugin-in-L, plugin-in-R] capture 2
```
and I don't see an obvious UI action to cycle from that config to e.g. a stereo setup, except for doing channels separately:
```
[plugin-in-L ] capture 1
[plugin-in-R ] capture 2
```

Perhaps you intend to swap source and target in the list. The we could visualize it as following:
```
[ capture-2 ] Plugin in L
[ capture-2 ] Plugin in R
```
but, even in that case, you'd need an action on both to go from mono to stereo.

The only simpler alternative the comes to mind is to simply list a few preconfigured presets, that we guess based on the number of capture channels and FX inputs, e.g. (assuming 2 capture channels and 2 FX inputs):
```
Stereo: capture-1, capture-2
Dual mono: capture-1
Dual mono: capture-2
```
which would indeed make it impossible to swap L and R. And we could still support 4-input capture cards quite cleanly without much extra work:
```
Stereo: capture-1, capture-2
Stereo: capture-3, capture-4
Dual mono: capture-1
Dual mono: capture-2
Dual mono: capture-3
Dual mono: capture-4
```

As to the use case for swapping L and R, it wasn't a use case I directly had in mind, but I've certainly had it in the past, where L and R were simply connected the wrong way, and fixing that without plugging cables can be handy. Specifically on zynthian, I expect 2 input channels to be more likely used for 2 instruments though, e.g. guitar and microphone.

@riban-bw Indeed there already is an LV2 plugin that does this quite well (Calf Stereo Tools), and I use it already under the current feature set. However, when compared to a simple Jack swap, it adds unnecessary CPU and memory load, both of which are of limited supply on the RaspPi ;-)

jypma:
I've created https://github.com/zynthian/zynthian-ui/pull/254 with a "simple UI" idea having a few mono/stereo choices.

jofemodo:
It's very simple. When you access the Audio Input for a given layer you should see that:

[ ] System: Capture 1
[ ] System: Capture 2

Simple checks. And that's all. From my POW, nothing more is needed.
The autoconnector will do its task depending of the "root" input plugin being Stereo or Mono.

+ If both inputs are selected and the root input plugin is stereo, it will connect both captures to each input and the sound will be stereo.

+ If both captures are selected, but the root input plugin is mono, it will connect both captures to the mono input and the sound will be down-mixed to mono.

+ If only one capture is selected and the input plugins is stereo, it will connect the selected capture to both inputs.

+ etc.

Simpler is better.

PR #254, opened 2020-06-16
Configure input capture channels for a layer

jypma:
This commit adds the ability to configure which capture inputs route to
the top input on an effects layer.

Internally, this is stored in a "audio_in" dict that precisely specifies
which capture inputs are connected to which LV2 input(s).

To the user, a simplified list of mono / stereo choices are provided,
depending on the number of capture inputs and number of LV2 inputs.

Supersedes #253 and implements the "simple UI" idea from there.

**TODO**
- [ ] Load and save `audio_in` in snapshot state, once agreed on the format
- [ ] Remove extraneous debugging lines and exception catching

### DIFF
--- a/zynautoconnect/zynthian_autoconnect.py
+++ b/zynautoconnect/zynthian_autoconnect.py
@@ -38,7 +38,7 @@ from zyngui import zynthian_gui_config
 # Configure logging
 #-------------------------------------------------------------------------------
 
-log_level = logging.WARNING
+log_level = logging.DEBUG
 
 logger=logging.getLogger(__name__)
 logger.setLevel(log_level)
@@ -408,7 +408,6 @@ def midi_autoconnect(force=False):
 	#Release Mutex Lock
 	release_lock()
 
-
 def audio_autoconnect(force=False):
 
 	if not force:
@@ -504,7 +503,7 @@ def audio_autoconnect(force=False):
 				pass
 
 	#Get System Capture ports => jack output ports!!
-	system_capture=jclient.get_ports(is_output=True, is_audio=True, is_physical=True)
+	system_capture=get_audio_capture_ports()
 	if len(system_capture)>0:
 
 		#Connect system capture to effect root layers ...
@@ -514,19 +513,33 @@ def audio_autoconnect(force=False):
 				continue
 
 			#Get Root Layer Input ports ...
-			rl_in=jclient.get_ports(rl.get_jackname(), is_input=True, is_audio=True)
+			rl_in=get_input_ports(rl)
 
-			#Connect System Capture to Root Layer ports
-			if len(rl_in)>0:
-				if len(rl_in)==1:
-					rl_in.append(rl_in[0])
-
-				try:
-					jclient.connect(system_capture[0],rl_in[0])
-					jclient.connect(system_capture[1],rl_in[1])
-				except:
-					pass
-
+			try:
+				for idx,src in enumerate(system_capture):
+					logging.debug("Trying {}".format(src))
+					disconnect = list(map(lambda x:x.name, rl_in))
+					for inputName in rl.audio_in.get(src.name, []):
+						disconnect.remove(inputName)
+						for dst in rl_in:
+							if dst.name == inputName:
+								logging.debug("{} -> {}".format(src, dst))
+								try:
+									jclient.connect(src, dst)
+								except:
+									pass
+								break
+					for inputName in disconnect:
+						for dst in rl_in:
+							if dst.name == inputName:
+								logging.debug("{} XX {}".format(src, dst))
+								try:
+									jclient.disconnect(src, dst)
+								except:
+									pass
+								break
+			except Exception as e:
+				logging.error("err", exc_info = e)
 
 		if zynthian_gui_config.midi_aubionotes_enabled:
 			#Get Aubio Input ports ...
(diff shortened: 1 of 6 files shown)


zynautoconnect/zynthian_autoconnect.py
```diff
@@ -38,7 +38,7 @@
 # Configure logging
 #-------------------------------------------------------------------------------
 
-log_level = logging.WARNING
```
jypma, line 41: Let's take this out before merging :-)

zynautoconnect/zynthian_autoconnect.py
```diff
@@ -510,19 +509,33 @@ def audio_autoconnect(force=False):
...
-					rl_in.append(rl_in[0])
-
-				try:
-					jclient.connect(system_capture[0],rl_in[0])
-					jclient.connect(system_capture[1],rl_in[1])
-				except:
-					pass
+			rl_in=get_input_ports(rl)
 
+			try:
```
jypma, line 514: the `try:` should also go here (was for debugging only).

However, I found that if `audio_autoconnect()` throws an exception, two problems occur:
- The exception doesn't make its way into the systemd journal logs (perhaps it's eaten or delayed?)
- The mutex is never released so the UI freezes

That's for a different PR of course; any ideas on how to approach this?

jofemodo:
There is only ONE input (root) plugin on every FX layer, so the only info we need to store is what capture channels are "enabled" for this layer: 1, 2, or both. That's all. The autoconnector will do its task and will connect everything accordingly to the number of input channels on the root input plugin.

jypma:
The decision to make the data model store routing decisions, rather than make them dynamically each time, was made purposefully. This enables us to evolve the UI separately from the data model, e.g. to implement more complex routing cases later. Routing logic need not be hidden from the user completely; a Zynthian user can be expected to know what "Mono" or "Stereo" means.

Please see the following screen shots of this branch:
mono
stereo
These are very close to your suggestion, yet more explicitly describe what is being routed.

If instead a multi-select is presented, and 4 capture inputs are available, it also becomes less obvious to the user that only 2 'X'es should be selected.

A second problem is that auto-detecting whether an LV2 is mono or stereo is hard to do reliably. Take for example the [LSP Sidechain Mono Compressor](https://lsp-plug.in/?page=manuals&section=sc_compressor_mono) (not currently in zynthian, but it's open source and popular). It has 2 audio inputs, but is not stereo.

jofemodo:
Hi @jypma !

Your solution is good. I like it (the first one too!). The problem with this one is that it doesn't scale if we decide to add support for more than 2 audio channels. And then we would need to add code to migrate snapshots ...

I would prefer a more "raw" approach, with the minimal coding complexity, keeping the same approach it's been used on the "Audio Out" routing. Having different approaches will create confusion ...

Regarding the problem of detecting whether an LV2 is mono or stereo, it's a marginal problem and it could be solved by using a simple "gain" element as root. Anyway, i think **this should be addressed** by LV2 specification...

jofemodo:
OK! Please, don't work on this until i release a basic implementation with the same approach than the current output routing. After that, we can open the discussion and see if more functionality is wanted/desirable and, in such a case, improve both routing tools (output and capture) at once.

jypma:
Sure thing! Thanks a lot for the discussion so far.

I actually did realize there's a case that my current "simple" setup doesn't cover: combining (adding) two capture channels into one mono plugin. The "X-es" approach would more naturally allow for that.

Though, how would we handle the UX of having 4 capture channels listed, but limiting the selection to only 2 of them?

Tying in to the coding discussion we were trying to get started, would this be a case to start experimenting with unit tests? `zynthian_autoconnect.py` is now going to handle lots of different cases, with already up to 5 indentation layers of looping/switching. Capturing a few expected use cases as tests could both help against regression and serve as documentation.

jofemodo:
> Sure thing! Thanks a lot for the discussion so far.

It was a pleasure ... and believe me, you make me think about future approaches on zynthian's routing. You put good ideas over the desk ... ;-)
>
> I actually did realize there's a case that my current "simple" setup doesn't cover: combining (adding) two capture channels into one mono plugin. The "X-es" approach would more naturally allow for that.

BTW, the basic "raw" implementation is commited. Perhaps you find interesting the autoconnect fragment that manage the capture connections. I tried to add support for more than 2 capture channels. And also extended this support to system output connections. But it must be tested with multi-channel cards...

https://github.com/zynthian/zynthian-ui/blob/7a4bfdf0dcec43b1d963e2a9b752fd16c802ed48/zynautoconnect/zynthian_autoconnect.py#L507

>
> Though, how would we handle the UX of having 4 capture channels listed, but limiting the selection to only 2 of them?

I didn't. If you select more than 2 channels, they will be down-mixed. But i limited the number of input ports on root plugins that will be connected to 2 ... so ...

+ if you have 4 channels and select all of them:
   - If the root FX-plugin is stereo: capture_1 & capture_3 are connected to port_1, capture_2 & capture_4 are connected to port_2.
   - If the root FX-plugin is mono: capture_1, capture_2, capture_3 & capture_4 are connected to port_1.


+ if you have 4 channels and select 3 of them:
   - If the root FX-plugin is stereo: capture_1 & capture_3 are connected to port_1, capture_2 is connected to port_2. OK! This is a little bit bizarre ... same than the selection!! Bizarre selection, bizarre result. But creative people love this kind of features!! ;-D
   - If the root FX-plugin is mono: capture_1, capture_2 & capture_3 are connected to port_1.

> Tying in to the coding discussion we were trying to get started, would this be a case to start experimenting with unit tests? zynthian_autoconnect.py is now going to handle lots of different cases, with already up to 5 indentation layers of looping/switching. Capturing a few expected use cases as tests could both help against regression and serve as documentation.

I really like the idea of unit tests, but it takes a lot of time to create and maintain!! Of course, you and everybody are very very welcome to add unit tests.

I've already added some unit tests to the MIDI-filter rule parser:

https://github.com/zynthian/zynthian-ui/blob/7a4bfdf0dcec43b1d963e2a9b752fd16c802ed48/zyngine/zynthian_midi_filter.py#L292

Perhaps this could be used as example. The "unittest" python module is really simple & clean to use ...

The best!

PR #258, opened 2020-06-21
Implement on-screen buttons

steveb:
This change shows 4 buttons across the bottom of the screen when
enable_touch_widgets is true (this is now option "Enable Touch Widgets" in the
webconf User Interface page)

Short/firm/long pushes are implemented, the button order for now is
Layer, Snapshot, Back, Select

### DIFF
--- a/zyngui/zynthian_gui_base.py
+++ b/zyngui/zynthian_gui_base.py
@@ -103,6 +103,54 @@ class zynthian_gui_base:
 		# Configure Topbar's Frame column widths
 		#self.tb_frame.grid_columnconfigure(0, minsize=self.path_canvas_width)
 
+		self.button_push_ts = 0
+
+	def init_touchbar(self):
+		# Touchbar frame
+		if not zynthian_gui_config.enable_touch_widgets:
+			return
+
+		self.touch_frame = tkinter.Frame(self.main_frame,
+		width=zynthian_gui_config.display_width,
+		height=zynthian_gui_config.touchbar_height,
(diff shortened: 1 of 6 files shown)


jofemodo:
Hi steve!
I just merged your branch into a new development branch on zynthian repo, so everybody can test it easily changing the branch from the webconf tool.

steveb:
Sweet thanks, I'll make any further PRs there

jofemodo:
Hi @steveb !

You have write access to the  zynthian:steveb_buttonbar development branch, so please, feel free to fix the problems.

Thanks!

riban-bw:
I think the init_buttonbar() call should be within zynthian_gui_base class rather than having to call it in every panel derived from that class. I keep finding panels that do not show the buttons. If these buttons are supposed to replace the hardware buttons then one might expect them to be omnipresent!

steveb:
> I think the init_buttonbar() call should be within zynthian_gui_base class rather than having to call it in every panel derived from that class. I keep finding panels that do not show the buttons. If these buttons are supposed to replace the hardware buttons then one might expect them to be omnipresent!

The only reason I did that was assumptions about the grid layout being filled up from top to bottom, it may be able to be done in the base class the way it is implemented now.

riban-bw:
Grid is only filled top to bottom if you do not specify a row. (Check out docs - it is inconsistent between rows and columns.) If you specify a row (as you have) then that column is populated, collapsing empty rows.

riban-bw:
This is implemented so closing.

PR #259, opened 2020-06-22
buttonbar tweaks

steveb:
- make height 1/6 of screen height instead of 1/5
- add missing datetime import
- reorder buttons, make labels capital
- attempt to get buttons to claim full height of bar (don't think it worked)


jofemodo:
Fantastic! ;-)

PR #260, opened 2020-06-25
Fix MIDI learning for value-based controllers

jypma:
Commit b49696d broke MIDI learning of value-based controllers, probably
because of a typo. Trying to MIDI learn a value-based controller results
in this "error" message in the log:
```
DEBUG:zynthian_controller: 'zynthian_controller' object has no attribute 'get_label2index'
```
This commit seems to fix that, by replacing the method call with what I
think was intended.


jofemodo:
Uoooohhh! Nice bug catch!! ;-)

Thanks a lot!

PR #262, opened 2020-07-02
Try to add setBfree manuals on consecutive MIDI channels

riban-bw:
Old behaviour: Adding setBfree will load the first manual to the requested MIDI channel then load subsequent manuals to the first free channels which may be lower, non-consecutive channels, e.g. MIDI channels: 5, 2, 3.
This patch attempts to load each manual of the setBfree engine on consecutive MIDI channels, falling back  to old behaviour of first available MIDI channel if consecutive channels are not available.

### DIFF
--- a/zyngine/zynthian_engine_setbfree.py
+++ b/zyngine/zynthian_engine_setbfree.py
@@ -272,7 +272,10 @@ class zynthian_engine_setbfree(zynthian_engine):
 			if self.manuals_config[4][0]:
 				try:
 					# Adding Lower Manual Layer
-					midi_chans[1] = free_chans.pop(0)
+					if midi_chans[0] + 1 in free_chans:
+						midi_chans[1] = midi_chans[0] + 1
+					else:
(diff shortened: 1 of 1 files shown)


jofemodo:
It seems OK!
Thanks a lot!

riban-bw:
I wonder if we should tweak further to try to populate any available _higher_ channels, e.g. if channels 5, 7 & 8 are populated and you try to add setBfree to channel 4, currently the first manual is added to channel 4 then the next two are added to 1 & 2. Maybe they should be added to channels 4, 6, 9? An alternative would be to constrain to only consecutive channels so in the above example only channels 1, 2, 9, 10, 11, 12, 13 and 14 would be offered to add setBfree to?

jofemodo:
Totally agree. I implement a new method "get_next_free_midi_channel()".
BTW i found and fix a bug that was breaking the keyboard split.

riban-bw:
You are a hero (who should avoid being distracted by such trivial requests :-) ).

PR #265, opened 2020-07-04
Feature/stepseq

riban-bw:
Merged master and tweaked to get on-screen buttons working. I have also removed "alpha" from the title in the main screen because I think we are past that stage. We could use "beta" if you want.
I suggest we merge this with master so that users can benefit from the significant changes since the version currently in master and to reduce the support overhead of users switching between incompatible versions.


zyngui/zynthian_gui_fileselector.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#******************************************************************************
+# ZYNTHIAN PROJECT: Zynthian GUI
+# 
+# Zynthian GUI Step-Sequencer Class
```
jofemodo, line 6: Mmmmm ... i think we could choose a better description for this piece of code ;-)

zyngui/zynthian_gui_base.py
```diff
@@ -412,26 +412,4 @@ def cb_topbar(self,event):
 		self.zyngui.zynswitch_defered('S',1)
 
 
-	def cb_keybinding(self, event):
```
jofemodo, line 415: I see you moved this to zynthian_gui and captured events from the top window.... this is completely OK for me ;-)

zyngui/zynthian_gui_fileselector.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#******************************************************************************
+# ZYNTHIAN PROJECT: Zynthian GUI
+# 
+# Zynthian GUI Step-Sequencer Class
+# 
+# Copyright (C) 2015-2020 Fernando Moyano <jofemodo@zynthian.org>
```
jofemodo, line 8: I can't remember to have contributed to this piece of code, but if you mean i "inspired" the code in some way, you can leave the line to rest  ;-)

zyngui/zynthian_gui_rename.py
```diff
@@ -0,0 +1,248 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#******************************************************************************
+# ZYNTHIAN PROJECT: Zynthian GUI
+# 
+# Zynthian GUI Step-Sequencer Class
```
jofemodo, line 6: Same applies here ;-)

zyngui/zynthian_gui_rename.py
```diff
@@ -0,0 +1,248 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#******************************************************************************
+# ZYNTHIAN PROJECT: Zynthian GUI
+# 
+# Zynthian GUI Step-Sequencer Class
+# 
+# Copyright (C) 2015-2020 Fernando Moyano <jofemodo@zynthian.org>
```
jofemodo, line 8: And here, of course ...

zyngui/zynthian_gui_base.py
riban-bw: Yes - we want to have consistent key capture across the application and putting this at the top level allows both a more consistent implementation and more uniform access to the binding.

riban-bw:
I have moved the constants back into the sequencer codebase but this does introduce a circular import. I don't think this is an issue because I don't actually use any functions or class variables, just these constants in one direction. I have made changes to address issues highlighted by this round of code review.

I had forgotten that I had added the _technology preview_ of on-screen keyboard and file selector but if you are happy to include this then we can devise a way to reuse in other modules.

PR #267, opened 2020-07-07
Handle differences between old and new mod-ui API

steveb:
This change should have no effect on the current forked zynthian mod-ui[1]
but it will also work with the zynthian changes applied to the
current upstream master mod-ui[2].

This change makes https://github.com/zynthian/zynthian-sys/issues/55 easier as
zynthian-ui will work with old and new mod-ui.

[1] https://github.com/zynthian/mod-ui
[2] https://github.com/steveb/mod-ui/tree/zyn-mod-merge


guysoft:
I am not authorized to merge this. @jofemodo ?

jofemodo:
OK! It's merged now ...

PR #270, opened 2020-07-17
Set pianoteq enginerate based on samplerate

### LINKED ISSUE #220: Controls don't always update when switching ZS3 presets on FX chain
I have an FX chain with 5 effects layered (see attached below). I have created 2 ZS3 sub-presets, which change parameters in the "Cabinet" effect.

When I send Program Change messages (PC1 and PC2), the LV2 plugins update fine, and the sound changes. However, the on-screen controls don't always update. Sometimes they do, sometimes they don't. Sometimes they do if I send PC2 two times. 

It seems to be a race condition on the control layer itself only; a bit of debugging showed that everything from picking up the values from the ZS3 preset, to sending that down into `jalv`, works fine. Somehow the UI just doesn't pick up the change.

It's not "just" a redraw issue either: when the bug occurs, the knob "really" is at the old position, as turning it will not make a sudden jump.

riban-bw:
Fixes https://github.com/zynthian/zynthian-issue-tracking/issues/220.

### DIFF
--- a/zyngine/zynthian_engine_pianoteq.py
+++ b/zyngine/zynthian_engine_pianoteq.py
@@ -35,7 +35,6 @@ from collections import defaultdict
 from os.path import isfile,isdir,join
 from xml.etree import ElementTree
 from json import JSONEncoder, JSONDecoder
-import jack
 
 from . import zynthian_engine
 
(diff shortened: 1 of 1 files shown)


zyngine/zynthian_engine_pianoteq.py
```diff
@@ -211,12 +212,18 @@ def fix_pianoteq_config():
 
 PIANOTEQ_CONFIG_FILE =  PIANOTEQ_CONFIG_DIR + "/" + PIANOTEQ_CONFIG_FILENAME
 
+jack_client = jack.Client("zynpianoteq")
```
jofemodo, line 215: I just added 2 functions to the main GUI class:

+ get_jackd_samplerate()
+ get_jackd_blocksize()

I think it's better to have this info easily available for all engines (zynaddsubfx, etc.)

Please, modify the code for using the new functions like:

    PIANOTEQ_SAMPLERATE = self.zyngui.get_jackd_samplerate()

zyngine/zynthian_engine_pianoteq.py
```diff
@@ -134,19 +135,19 @@ def fix_pianoteq_config():
 				logging.debug("Fixing Audio Setup")
 				for devicesetup in audio_setup_node.iter('DEVICESETUP'):
 					devicesetup.set('deviceType','JACK')
-					devicesetup.set('audioOutputDeviceName','Auto-connect ON')
-					devicesetup.set('audioInputDeviceName','Auto-connect ON')
-					devicesetup.set('audioDeviceRate','44100')
+					devicesetup.set('audioOutputDeviceName','Auto-connect OFF')
```
jofemodo, line 138: This is a good point. It should prevent strange effects when loading snapshots with pianoteq+effects layers.

zyngine/zynthian_engine_pianoteq.py
jofemodo: OK! I've already changed the code.

PR #272, opened 2020-07-22
Adds UP/DOWN buttons

riban-bw:
This feature has been requested by several users and having used the onscreen buttons for a while I concur.


riban-bw:
Assigning back to @steveb to merge updates from testing branch and resolve feedback.

steveb:
> Don't you think we should better take profit of the current "zynthian logo" space, at top-right?
> Come on!!! ;-)

That only works on screens which have the logo, it would be nice if the up/down buttons are in a stable position for all screens.

steveb:
Closing so we can do the reviews in https://github.com/zynthian/zynthian-ui/pull/363

PR #304, opened 2020-11-29
Fix Celtic Harp group ID

josephfh:
Problem: Celtic harp was showing in 'demo items' when bought and licensed

Fix: change the group ID to the correct value


jofemodo:
Merged! Thanks!

PR #308, opened 2021-01-04
Feature/touchscreen calibration

riban-bw:
This now works as requested.

- Screen shows target crosshairs that must be touched and released
- First target does no calibration and only detects the touch interface
- Display type shown after first touch
- Consequent touches are measured on release, allowing drag to target for more accuracy
- After second touch the calibration is written to two files (I disagree!)
- Each target appears white and changes to red when screen touched
- Touches too close together are ignored (half a screen)
- A timeout of 15s after last touch or release will close screen without saving configuration
- BACK button will close screen without saving configuration


jofemodo:
OK! Let's go with it and we will add the multi-device in a second pass

PR #309, opened 2021-01-07
VNC setting on admin menu

Pastitas:
As i explained in the forums, I want to develop some things for zynthian and this is kind of a PoC.

Disabling VNC proves advantageous as [it can be considered a security risk](https://security.stackexchange.com/questions/124958/how-do-i-assess-and-mitigate-the-security-risks-of-a-vnc-tool), since misconfiguration might lead to an increased attack surface. 

This merge introduces a menu option in the admin menu to enable/disable the VNC service (just runs systemctl start and enable and stop and disable for the services and checks for the service status).

I don’t know if maybe this is more useful if i add an install VNC option first if the service is not installed, i was just trying to get acquainted with how the menu system works.

This does persist between reboots but it does so by using systemd.

### DIFF
--- a/zynconf/zynthian_config.py
+++ b/zynconf/zynthian_config.py
@@ -482,6 +482,18 @@ def is_service_active(service):
 		return True
 	else:
 		return False
+		
+def is_service_enabled(service):
+	cmd="systemctl is-enabled %s" % service
+	try:
+		result=check_output(cmd, shell=True).decode('utf-8','ignore')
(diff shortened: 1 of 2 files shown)


jofemodo:
Hi Pastitas!

RealVNC is not free software and it's not available from the Raspbian repositories. I would prefer to use TigerVNC that is fully free software and it's available from the raspbian repos.

I'm thinking on it ...

Thanks!

PR #312, opened 2021-01-31
Fixes failure to calibrate screen if ambiguous devices connected

riban-bw:
I have added exception handling to the xinput function to allow the calibration to complete or fail gracefully if an error occurs. A user found that their multifunction keyboard / touchpad was triggering xinput to fail due to abmiguity of the device names. This fix allows calibration to complete.


jofemodo:
Done!

Issue #344, opened 2021-02-26
Ability to move patterns in arranger view


riban-bw:
**Is your feature request related to a problem? Please describe.**
When in arranger it feels natural to click and drag patterns but currently you need to click to remove then click to add in new position.

**Describe the solution you'd like**
With mouse / touchscreen click and drag should allow moving a pattern to another location in the timeline or another track.
With encoders it should be possible to mark a track then move it with the up/down/left/right encoders.

**Additional context**
This is very obvious when using a mouse or touchscreen but would be advantageous with encoders too.

riban-bw:
Added click and drag to move patterns in Arranger to development. Have not added ability to move with encoders. Not sure how best to mark a pattern to be moved and am not convinced it provides much benefit. The click and drag seems to work quite well and is intuitive. Currently moves a wireframe which may benefit from being filled - comments?

riban-bw:
Done awaiting merge with testing branch.

Issue #343, opened 2021-02-26
Sequence with multiple tracks play all notes fast, not correctly timed


riban-bw:
**Describe the bug**
Sequences that have more than one track playback all the notes of the sequence immediately rather than at the correct scheduled time.

**To Reproduce**
1. Go to ZynPad
2. Add a pattern to the first pad / sequence
3. In arranger add a second track
4. Play sequence from zynpad

**Expected behaviour**
Pattern plays correctly.

**Actual behaviour**
All notes in pattern play immediately.

**Setup:**
HARDWARE
Raspberry Pi 4 Model B Rev 1.4
Soundcard: HifiBerry DAC+ ADC PRO
Display: ZynScreen 3.5 (v1)
Wiring: MCP23017_ZynScreen
GPIO Expander: MCP23017

SYSTEM
Raspbian GNU/Linux 10 (buster)
Build Date: 2020-09-05
Memory: 21% (426M/1939M)
SD Card: 36% (9.4G/28G)
Temperature: 48.0ºC

SOFTWARE
zyncoder: master (ef24e0e)
zynthian-ui: master (acf5297)
zynthian-sys: master (6b2a68f)
zynthian-data: master (2028b5d)
zynthian-webconf: master (d57e6e2)

**Additional context**
This is an significant flaw that should be fixed urgently.

Issue #338, opened 2021-02-20
ZynSeq: Clicking on the label of a pattern in arranger does not remove pattern


riban-bw:
**Describe the bug**
Clicking on a pattern in arranger removes the pattern from the track but this fails if you click on the pattern's label.

**To Reproduce**
1. Go to Arranger
2. Add pattern to a track
3. Click on the white text of the pattern number

**Expected behaviour**
Pattern is removed.

**Actual behaviour**
Pattern is not removed.

**Setup:**
HARDWARE
Raspberry Pi 3 Model B Plus Rev 1.3
Soundcard: HifiBerry DAC+ ADC PRO
Display: ZynScreen 3.5 (v1)
Wiring: MCP23017_ZynScreen
GPIO Expander: MCP23017

SYSTEM
Raspbian GNU/Linux 10 (buster)
Build Date: 2020-05-27
Memory: 10% (94M/926M)
SD Card: 39% (11G/29G)
Temperature: 49.4ºC

SOFTWARE
zyncoder: master (dcb627e)
zynthian-ui: feature/stepseq (88d1458)
zynthian-sys: master (6d93662)
zynthian-data: master (2028b5d)
zynthian-webconf: master (45bcb08)

**Additional context**
Need to add mouse handler to the text (or add tag to text that is already handled).

riban-bw:
Fixed in development branch. Pull request submitted to merge to testing branch.

Issue #336, opened 2021-02-20
ZynSeq: Press and hold pad to access editor also triggers pad to start


riban-bw:
**Describe the bug**
When using a touchscreen the way to access pattern editor is by press and hold a pad in zynpad. This starts the sequence playing which is undesirable.

**To Reproduce**
1. Go to ZynPad
2. Press and hold a pad

**Expected behaviour**
Pattern editor opens. Pad does not play.

**Actual behaviour**
Pattern editor opens. Pad starts to play.

**Setup:**
HARDWARE
Raspberry Pi 3 Model B Plus Rev 1.3
Soundcard: HifiBerry DAC+ ADC PRO
Display: ZynScreen 3.5 (v1)
Wiring: MCP23017_ZynScreen
GPIO Expander: MCP23017

SYSTEM
Raspbian GNU/Linux 10 (buster)
Build Date: 2020-05-27
Memory: 10% (94M/926M)
SD Card: 39% (11G/29G)
Temperature: 49.4ºC

SOFTWARE
zyncoder: master (dcb627e)
zynthian-ui: feature/stepseq (88d1458)
zynthian-sys: master (6d93662)
zynthian-data: master (2028b5d)
zynthian-webconf: master (45bcb08)

**Additional context**
We want the pad to start promptly hence it triggers on press rather than release. It is not obvious how best to resolve this. We could use the action of opening the pattern editor to stop the sequence playing. This kind of makes sense because you generally don't want a pattern playing externally whilst editing it in pattern editor.

PR #345, opened 2021-02-26
StepSeq Sprint 2021.02.1

riban-bw:
This PR applies the fixes scheduled for Sprint 2020.02.1:

Sequence with multiple tracks play all notes fast, not correctly timed
    https://github.com/zynthian/zynthian-ui/issues/343
Use pattern index as default name for sequence.
Select pad in zynpad then open arranger - selects Nth track but should select Nth sequence.
Add ability to clear OSK input field.
"DELETE" does not fit button in OSK.
Renaming sequences feels more appropriate from zynpad.
Limit length of QWERTY keyboard input, e.g. don't allow more thatn 16 characters for pad names.
Show group on pad near / on play state icon.
Button images render poorly.
    https://github.com/zynthian/zynthian-issue-tracking/issues/353
Use long press in menu to reset value or set to default, e.g. disable MIDI trigger note.
    https://github.com/zynthian/zynthian-issue-tracking/issues/318
Should warn before destructive actions: e.g. reducing beats in pattern if note will be lost.
    https://github.com/zynthian/zynthian-issue-tracking/issues/320
Should use standard confirmation dialog for destuctive operations or yes/no questions like clear pattern. 
    https://github.com/zynthian/zynthian-issue-tracking/issues/315
Click on pattern label does not trigger toggle.
    https://github.com/zynthian/zynthian-issue-tracking/issues/333
Adjusting note duration is not affecting selected note.
    https://github.com/zynthian/zynthian-issue-tracking/issues/345
Note length should be indicated if it exceeds beyond end of display, e.g. +2
(Previous bug: Changing pattern length does not update end of pattern note length text, e.g. +2)
    https://github.com/zynthian/zynthian-issue-tracking/issues/344
Pattern editor changing tonic does not work.
    https://github.com/zynthian/zynthian-issue-tracking/issues/343
Unable to exit ZynSeq with touchscreen if Touch Widgets disabled. Need to show exit / back button even if touch widgets disabled.
    https://github.com/zynthian/zynthian-issue-tracking/issues/334
It is easy to change bank without noticing. Suggest adding "Bx" prefix to each pad to indicate its bank, e.g. B1A1.
Name sequences using qwerty keyboard.
    https://github.com/zynthian/zynthian-issue-tracking/issues/316
Improve SMF import
    https://github.com/zynthian/zynthian-issue-tracking/issues/323
Changing steps per beat will move notes between beats, e.g. a note on step 1 of beat 2 may move to step 5 of beat 1 when changing from 4 to 6 steps per beat.
    https://github.com/zynthian/zynthian-issue-tracking/issues/313
Changing grid size in zynpad moves existing sequences to other groups. Ideally it would reorder so that they stayed in the same group.
    https://github.com/zynthian/zynthian-issue-tracking/issues/317
Show vertical bar lines in arranger. A timebase track with times shown numerically. This may be beneficial too to show bar number, especially when scrolled.
    https://github.com/zynthian/zynthian-issue-tracking/issues/319
Cannot set trigger channel to "None".
    https://github.com/zynthian/zynthian-issue-tracking/issues/338
Pattern editor after copying pattern the source pattern for subsequent copies is not changed to current (new) pattern.
    https://github.com/zynthian/zynthian-issue-tracking/issues/348
Press and hold pad to access editor also triggers pad to start. Could stop sequence if opening editor.
    https://github.com/zynthian/zynthian-issue-tracking/issues/336
After copy pattern set arranger pattern to the new pattern

Note that issue zynthian/zynthian-ui/issues/343 is a break-fix so should be applied as soon as validated. The others are issues identified for this sprint or dependancies for this sprint.


jofemodo:
Hi @riban-bw !

I have some questions/comments: 

+ Initially, sequence numbers are 1-16 for the first bank, 17-32 for the second bank and so on. If i change the grid size of bank 1 to 6x6, then i have sequences 1-36 on first bank, but bank 2 still have 16-32. I see they are different ones, but it's strange. If this sequence numbers (or names) can be repeated across banks, why the initial distribution?

+ When editing a pattern, the title says "pattern x (pad y)", but any of these numbers match the sequence numbers that you can read on the pads. I mean, if the pad number is not displayed on pads, why to display it here?

+ When i access the arranger i see a title like "Bank 1 J22-1 (10)". Woaooo!! J is the group, 22 is the sequence and the 10 is the MIDI channel, but what is the "-1" following the 22?

+ When the popup menu is displayed, there is a ugly "X" that shouldn't be visible unless "touch widgets" are enabled. Is it?

Sorry for not merging tonight, but i would like to understand well and test a little bit more. I'm really tired now, so i will continue tomorrow.

Thanks for your patient!

riban-bw:
Hi @jofemodo 

No worries - all reasonable questions and this is the point of pull-requests. I appreciate the peer review.

Initially there are no sequences and no banks. When a new bank is selected a grid of 4x4 sequences are created and give default names. Some time ago you suggested we have different / unique names for each sequence to ease distinguising them> I was less keen but subsequently found it easy to misread which bank I was on so started to redress this. I added the ability to name sequences which allows a user to call any sequence anything they like, including calling them the same thing! We have a _simple_ mode which we want to be as automated as possible. When working in ZynPad the default is to create sequences with a single pattern. Each sequence is unique and each new sequence is assigned a unique pattern. (This can be changed later.) The default name for the sequences (pads) in ZynPad is the number of its initial pattern. This is quite a good implementation for users who only use simple mode. Every sequence has one pattern and the name of the sequence is the same as the name of the pattern. Once a user starts to use arranger then they relinquish the simple / automated naming convention and can start to name things to suit their workflows. Hopefully that excess of information answers your first question.

When the pattern editor is opened from ZynPad a check is performed to see whether the pattern satisfies the conditions of a simple mode pad, i.e. is this the only pattern in the sequence and at the start of the sequence. If so the pattern editor is opened and shows details about the pattern and pad. If not then Arranger is opened. Opening pattern editor from arranger does not do this check. If the pattern editor has been opened from ZynPad then the title shows "Pattern xx (Pad yy)". xx is the actual pattern index. yy is the offset of the pad within the grid. This may not be the same as the sequence (pad) name if the grid size has been changed. I did not consider the impact on this when I added the ability to rename sequences. I guess we should actually show the sequence name rather than the pad index. I can make that change.

I thought I had described the track titles but maybe not fully or not in an obvious place. (I did not want to update wiki until after the code drops.) The title in Arranger shows the bank, group, sequence, track and channel. the track within the sequence is shown after the hyphen. It is also shown in the track's title bar for tracks above 1. This is to allow the sequence name more space which is shown where track 1 would be. It may not be intuitive when you only have single track sequences. I am not sure whether there is benefit in hiding the track numeber in the title for single track sequences and showing it for multitrack sequences. What do you think?

Beauty is in the eye of the beholder. I changed the icons to make them less ugly. The X is large but simple. This is a change I made in this sprint to fix the inability to exit step sequencer with a touchscreen. A user may not want to use touch widgets or may have not enabled them so they are given a method of exiting the screen by showing the menu and selecting the X. Previously this was a back arrow. I could change it for a different icon but despite your aversion to seeing any icons here I do think it advantageous. The titlebar is not used at that point so is available and there has been a common query in the past about how we exit screens. I think this makes it more intuitive. (I considered using the hyper-cryptic, need-to-know implementation of tapping the topbar but my user thinktank didn't like it!!!) So this is actually by design. In previous version the X was hidden if touchwidgets were disabled but I found this had significant impact on user experience.

I have started work (actually quite advanced) on next sprint but have avoided pushing because I think it will interfere with the PR. Maybe I should create a new branch for each sprint (yes I think I should) to avoid such interference. I will do that and then apply the change for sequence name in pattern editor to this branch.

riban-bw:
Changed pattern editor to show name rather than pad offset. I also changed it to say "Seq:" rather than "Pad". Not sure which is better.

jofemodo:
Replaced by https://github.com/zynthian/zynthian-ui/pull/353

PR #353, opened 2021-03-12
Sprint 2021.02.27

riban-bw:
Fixes these issues:

- zynthian/zynthian-ui/issues/344 Move patterns in Arranger
- zynthian/zynthian-ui/issues/346 Convert onscreen keyboard to be compatible with other Zynthian modules
- zynthian/zynthian-ui/issues/331 Ability to play from Arranger


jofemodo:
Moving the snapshot encoder cause the tempo to be displayed but it doesn't change. BPM is fixed to 0.0 and it plays reaaaaaaaaaaaaaaaallyyyyyyyy sloooooooooooooooooooooooooooooooooooooooow ;-)

jofemodo:
When the top-left menu is displayed, the big "X" is still there, on the top-left corner, although "Touch Widgets" option is disabled on webconf.

jofemodo:
I can't figure how to fine-change BPM with the rotary. Is it possible without using the touch screen?
I would use the rotary acceleration feature for that. If you agree, i could implement it on the zynseq.

jofemodo:
It's merged!!
I'm still exploring everything, but it work nicely. Very good work!!

riban-bw:
Cool! I will update the wiki to match recent changes.

riban-bw:
I have updated wiki but each time I try to replace an image it just reloads the old image thinking it is the same.

I have pushed a fix for tempo. I had previously checked if touch widgets were enabled and hid the tempo menu item but now that it is used for fine adjustment we need it so I have removed the check. Please merge this extra fix.

The 'X' is there by design to allow users to exit the screen using touchscreen even if they don't have touch widgets enabled. If you don't like the look I could make it more subtle or pretty.

jofemodo:
Hi @riban-bw !

The wiki issue with thumbnails seems to be solved. A full morning wasted to solve a rare locale/permission error that finally didn't get to fully understand, but i managed to get it working.

Regarding the "X", i don't like it. If you have rotaries and switches, you don't needed at all. If you don't have, then you should enable the touch widgets. In fact, we could enable this by default when "dummies" wiring is selected.

Regards,

jofemodo:
Regarding the tempo fine-change, we have to think something better. It's quite ugly having 2 ways of changing the tempo:

+ Using the menu option, then touch widgets are shown and you use it for coarse-change, while rotary do fine-change. No way of fine-adjustment using only touch?
+ Moving the SS rotary, then no touch widgets are shown and only coarse adjustments is possible.

jofemodo:
Hi @riban-bw !

I just commited to master 2 little changes:

+ Enable rotary acceleration on StepSequencer, and make it work with BPM fine change. It works really smoothly!!

+ Hide all touchwidgets when the config flag "enable_touch_widgets" is disable. We created this flag for something. People disabling touch widgets don't want to see touch widgets. People enabling touch widgets, clearly want the touch widgets. It's very simple. If you think we need a middle point, OK, let's replace the binary flag for something like "touch widgets level", with 3 values: "NONE", "SOME, "ALL". This would be OK ;-)

Cheers!

riban-bw:
You disabled rotary encoder acceleration on StepSequencer because it was interfering with user workflow so enabling it across the whole module may prove suboptimal. I checked the tempo control and it works quite well (although it doesn't suit the requester's desire for even smaller steps). I will add a modification to the encoder registration in stepsequencer to allow enable / disable of acceleration.

Regarding touchwidgets - I had an actual problem being stuck in the stepsequencer with just a touchscreen which I felt was poor user experience hence enabled the back button in the menu. (I think use of the topbar as a hidden back button is poor user experience.) If you feel this is inapproriate and that users must explicitly enable touchscreen widgets if using touchscreen then so be it. I will just accept your judgement (and let you take any flak for it :wink:).

Unfortunately you broke stepsequencer with this change. I will add patches to fix this break and allow selective enabling of acceleration.

riban-bw:
Actually... I HATE ZYNCODERS!!!!!! Bloody steps of 4 when you enable acceleration. It does my head in!

jofemodo:
Yes, acceleration is not comfortable for small value ranges. It should be enable when the number of ticks (choosable values in the range) is bigger than 50-100. It's what i do on the controller class. Please, let me patch the code for selective enabling the acceleration only when needed.

Anyway, it's not true that steps are 4. Step size ranges from 1 to 8, depending of the rotary speed. When you move slowly, it step 1 by 1. This is with the original zyncoder library. I've not reviewed the i2c implementation.

Regarding the touch widgets, from my POV, if we have a flag for disabling the touch widgets, it must disable the touch widgets. It's a question of coherence. In the other hand, if you think we should have a middle point where some touch widgets are displayed and others no, let's implement it. No fights, no worries ;-)

Regards!

jofemodo:
Hi @riban-bw !

I just patched the code so the "dynamic step" feature can be enable/disable for each encoder. It's enable by default on the snapshot encoder and disable for the rest. I would like to enable this on the pattern editor cursors, specially when grid size is big.
What do you think?

Regards,

PR #359, opened 2021-03-17
Fix for zynstep encoders

riban-bw:
Adds option to accelerate encoders in zynstep.
Fixes zynseq touch widgets bug.

This is a critical fix. Master is currently broken.

### DIFF
--- a/zyngui/zynthian_gui_stepsequencer.py
+++ b/zyngui/zynthian_gui_stepsequencer.py
@@ -128,6 +128,7 @@ class zynthian_gui_stepsequencer(zynthian_gui_base.zynthian_gui_base):
 		self.zyngui = zynthian_gui_config.zyngui # Zynthian GUI configuration
 		self.bank = 1 # Currently displayed bank of sequences
 		self.layers = [None for i in range(16)] # Root layer indexed by MIDI channel
+		self.max_step = [1,1,1,1] # Maximum steps each encoder can change
 
 		#libseq.enableDebug(True)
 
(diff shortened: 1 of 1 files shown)


jofemodo:
Hi @riban-bw !!

Didn't you see my patch from this morning? Please test it:

https://github.com/zynthian/zynthian-ui/commit/e34cc5d52cb7dfd2f8872a416cbe7f38184063f3

Anyway, how can you say that master is broken? It's not broken at all, my darling . It works quite decently, yet without my latest patch. And currently it works perfectly OK. If this is not true for you, please, specify the problem.

BTW, are using the zyncoder i2c library? Perhaps the "dynamic step size" is not properly implemented there ... 

Cheers!

riban-bw:
I just saw your patch which shares similarities with the one I submitted as a PR to fix this.

Master is broken because there is no check for hiding the buttons hence fails when trying to select a menu option. This is fixed in my PR.

I am using standard zynthian hardware and zyncoder libraries. I have not patched i2c library for a while because that hardware is not currently connected to any of my devices.

zyngui/zynthian_gui_stepsequencer.py
```diff
@@ -403,8 +401,7 @@ def show_menu(self):
 			self.unregister_zyncoder(encoder)
 		self.register_switch(ENC_SELECT, self)
 		self.register_switch(ENC_BACK, self)
-#		if zynthian_gui_config.enable_touch_widgets:
-		if True:
+		if zynthian_gui_config.enable_touch_widgets:
```
jofemodo, line 404: You forgot to enable the line 416 too ;-)

zyngui/zynthian_gui_stepsequencer.py
```diff
@@ -1040,6 +1042,10 @@ def register_zyncoder(self, encoder, object):
 			pin_a=zynthian_gui_config.zyncoder_pin_a[encoder]
 			pin_b=zynthian_gui_config.zyncoder_pin_b[encoder]
 			zyncoder.setup_zyncoder(encoder, pin_a, pin_b, 0, 0, None, 64, 128, 0)
```
jofemodo, line 1044: The last parameter on zyncoder.setup_encoder is the step size. When 0 is passed, dynamic step size (acceleration) is enabled.

jofemodo (commented):
As this PR is conflicting with my latest commit, i prefered to cherry-pick your changes, re-enabling the missed touchwidget conditions, but keeping the current zyncoder code, that fully works.

PR #366, opened 2021-03-24
Fixes engines failing to start or show UI when toggling VNC

riban-bw:
Fixes #zynthian/zynthian-issue-tracking/issues/407.

- Detects if LV2 has UI during scan for new plugins
- Checks list of broken UIs that don't work on Zynthian
  - Helm only works on RPi4 so checks for that too
- Starts jalv.gtk if VNC server running forwarding display to VNC display
- Starts jalv.gtk if `ZYNTHIAN_X11_SSH` env var set and forwards display to ssh X11 display
- Starts jalv if UI or display not available
- Does not show generic UI if there is not an engine specific UI (many audio effects have generic looking displays defined fro modui)
- Requires user to search for new plugins and presets to benefit from changes

I would prefer to change the name of the vnc service to lose `@\:1` but that is subject of a different request. It will require changes to multiple repositories (and fixing current installations) so deferring for now.

I have tested this on official Zynthian hardware 3 & 4.1 with Raspberry Pi 3 and 4. It seems to operate as expected with many issues resolved. Please test further before merging to _testing_ branch.

### DIFF
--- a/zyngine/zynthian_engine.py
+++ b/zyngine/zynthian_engine.py
@@ -62,7 +62,6 @@ class zynthian_basic_engine:
 		self.proc_timeout = 20
 		self.proc_start_sleep = None
 		self.command = command
-		self.command_env = None
 		self.command_prompt = prompt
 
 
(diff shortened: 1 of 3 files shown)


jofemodo:
Hi @riban-bw !

It seems OK, but i couldn't test yet. I will do ASAP ...

Thanks!

jofemodo:
I've been testing but it doesn't solves the issue for me. I followed the steps from the description and it still fails.
BTW, I've detected a problem that could be related. Some plugins, when run with jalv.gtk, generate async error messages that could be confusing the parser. I've tried to remove/ignore the stderr output, but pexpect uses a "pty" (virtual terminal) and it's not so easy. Let me try a few things ;-)

jofemodo:
Merged, although it doesn't solve all cases. I keep open https://github.com/zynthian/zynthian-issue-tracking/issues/407

PR #378, opened 2021-04-22
Partial fix for memory leak - limits to 10% RAM usage

jofemodo:


zyngine/zynthian_lv2.py
```diff
@@ -293,7 +293,6 @@ def _generate_plugin_presets_cache(plugin):
 	# Get banks
 	banks = plugin.get_related(world.ns.presets.Bank)
 	for bank in banks:
-		world.load_resource(bank)
```
jofemodo, line 296: This is needed because the rdfs.label is not always available without loading the resource. It depends on how the bank has been defined on TTLs.

jofemodo:
Let's merge this to testing, but it must be carefully tested, specially the webconf's preset manager.

PR #385, opened 2021-05-28
Fixes #461 - stuck MIDI notes in SMF player.

riban-bw:
Was sending note-off for wrong channel.
Conversion of NOTE OFF to NOTE ON with zero velocity ignored channel.


jofemodo:
It works perfectly now ;-)

Thanks!

PR #389, opened 2021-06-30
Feature/747 indicate empty sequence

riban-bw:
This PR is to merge the development branch which implements feature request 747. A _stopped_ icon is shown on each zynpad when the associated sequence is stopped _and_ the sequence contains events. This is to give an indication to user of which pads contain events and which are empty which may help some user workflows.


jofemodo:
Merged!

PR #391, opened 2021-07-01
Bug/477 xruns

riban-bw:
Rationalises and fixes incorrect initialisation of zynseq library.


riban-bw:
Wait! Just found a possible bug. Testing...

ZynPad is not starting clock. I need to look at this some more.

PR #392, opened 2021-07-02
Bug/477 xruns

riban-bw:
Fixes xrun issue but rationalising zynseq library handling.
Also includes a small fix for zynpad not clearing play icon when stopped (if sequence is empty).


zynlibs/jackpeak/jackpeak.c
```diff
@@ -70,7 +70,7 @@ int initJackpeak() {
 
 	// Register the cleanup function to be called when program exits
 	// It raises a segmentation fault on exit, so it's disabled
-	//atexit(endJackpeak);
+	atexit(endJackpeak);
```
jofemodo, line 72: DId you check if the segmentation fault is still there when exiting the UI?

riban-bw: I have committed a change that removes the comment and actually runs the exit code. Segmentation fault is resolved.

jofemodo:
Merged!

PR #399, opened 2021-09-18
Midi bank change

Vlietland:
Solved the 403 error by replacing the fork. Implemented as follows:
When CC0 (MSB) is received, saved to flag var "bank_msb".
When CC32 (LSB) is received, bank changes, using the "bank_msb":
bank_msb==0 => system banks LSB (0-127)
bank_msb==1 => user bank LSB (0-127)
bank_msb==2 => external USB bank LSB (0-127)
bank_msb==3 => favourites bank, if it exist. In this case, lsb has not meaning.

Used following tests with database -> favorites, user & system banks and no external bank
MSB = 0 / LSB 1-x -> works
MSB = 1 / LSB 1-x -> works
MSB = 2 / LSB 0 -> throws invalid midi cc warning and keeps the current setting
MSB = 3 -> throws invalid midi cc warning since it can't load favorites in fluidsynth

Unrelated bugs: 
* Fluidsynth cannot select the favorites bank and throws an error when favorites bank is selected (see error log). Seems a separate issue. Don't know external bank ID, now named it "EXT". Probably needs updating.
* GUI is not updated in layer, bank and preset screen
* GUI is sometimes not updated in control screen

ps: Atom automatically removes spaces at the end of text, leading to some extra changes. Hope you don't mind :-)

### DIFF
--- a/zyngine/zynthian_layer.py
+++ b/zyngine/zynthian_layer.py
@@ -43,6 +43,11 @@ class zynthian_layer:
 		self.zyngui = zyngui
 		self.engine = engine
 		self.midi_chan = midi_chan
+		self.bank_msb = 0
+		self.favs_available = False
+		self.first_extbank_index = 0
+		self.first_userbank_index = 0
+		self.first_systembank_index = 0
 
 		self.jackname = None
 		self.audio_out = ["system:playback_1", "system:playback_2"]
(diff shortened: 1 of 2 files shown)


jofemodo:
This functionality is already implemented in the  “midi_bank_change" branch.
I used your previous PR code as a base and added the MSB functionality (including the range checks!).
I explained this on the commit message and gave you credit, but it's authored by me. I hope you don't feel annoyed. I hope to receive more PRs with good ideas (like this!!) from you in the future.

BTW, take a look to my implementation. It's more efficient, simple and avoid some problems that your implementation have. The key difference is it intercepts CC0 & CC32 in the upper code layer (zynthian_gui_layer). In fact, this is not a good place, but "zynthian_layer" class is not neither for several reasons. The problem is that we would need an extra code layer here, a "zynthian_chain" class, but it's not properly implemented. OK, the functionality is implemented but not in the right place. The "zynthian_chain" functionality is implemented inside "zynthian_gui_layer", what makes this class a big mess and one of the weaker points in the current architecture. It's the first candidate to refactorization and code-splitting.

This kind of subtleties are difficult to explain and it's the reason because i implemented the code myself. Now i see that it could be an error because you could feel annoyed and i really didn't pretend it. I simply wanted to implement it in the best way (given the current code base) and not wasting too much time trying to explain you all the details. Anyway, i would like to pay my fault with all the detailed explanations you want. You seem a good programmer and i would like to gain a collaborator for the cause ;-)

All the best,

Vlietland:
For me most important is that the product gets more robust. Currently the product has too many bugs. I'm fine with programming it all yourself unless it's buggy and I can't use it for serious composing. With all product I used (Kodi, Domoticz, Linux kernel, Debian, KDE, Lineage, Rosegarden and many many many more) Zythian is most buggy. That is what annoys me. So a reliable MVP with less features is fine for me.

Btw I pulled your code Friday but the problem was still there. No proper handling of midi. I therefor wrote the code myself. I will keep using my code until testing shows that the issues are really solved. 

Cheers!

jofemodo:
That's the key. We all want a more robust zynthian with a better architecture and code.
Anyway, you have to realize that your use-case is not the primary use-case, so it's normal you find more issues. But it's good that zynthian is tested with different use-cases, as it's the better way of pushing forward and finding bugs/problems.

Regarding your problems with my branch, you should be more explicit. It works for me, so please, tell me the steps for reproducing the error. 

Regards,

jofemodo:
Closed because it's already implemented. I would like to know why the implemented solution is not OK for you, @vlietland. If you tell me where the problem is or how to reproduce it ...

PR #402, opened 2021-10-14
Update zynthian_autoconnect.py

hippich:
Respect ZYNTHIAN_LOG_LEVEL env variable in `zynthian_autoconnect.py`


jofemodo:
OK! Let's try this. Anyway, if it's "too noisy", i will undo the change or look for alternatives.

Issue #401, opened 2021-10-01
Onscreen button long-press gets triggered when changing bottom row parameter


steveb:
To reproduce:
- Enable Enable Onscreen Buttons
- Navigate to a Control screen 
- Use touch screen control to change the bottom left parameter
- After using touch to change the parameter value for about 5 seconds, let go
Expected behaviour: Nothing
Actual behaviour: The BACK button Long-Back action is triggered and the screen changes

The same thing happens for the bottom right control and the SELECT button.

Its like the active area of the back button encroaches on the bottom left parameter control.

I have a touchscreen-only zynthian, so this is making it a bit hard to use.

steveb:
There is an argument for a control never handling button presses specifically. We could assume there will either be onscreen buttons along the bottom, or physical buttons.

At the very least, if onscreen buttons are enabled, button behavior can be disabled on the controls.

PR #403, opened 2021-10-16
Don't use controller as button in some cases

steveb:
This change reduces the situations where a controller will trigger a
button push:
- If onscreen buttons are enabled, never use controllers as buttons
- If the controller value has been changed during the push event,
  don't trigger a button on release

This makes it practical to use a touchscreen-only zynthian to change
the controller values without triggering an undesired button press.

This fixes issue #401

### DIFF
--- a/zyngui/zynthian_gui_controller.py
+++ b/zyngui/zynthian_gui_controller.py
@@ -2,13 +2,13 @@
 # -*- coding: utf-8 -*-
 #******************************************************************************
 # ZYNTHIAN PROJECT: Zynthian GUI
-# 
+#
 # Zynthian GUI Controller Class
-# 
+#
 # Copyright (C) 2015-2016 Fernando Moyano <jofemodo@zynthian.org>
 #
 #******************************************************************************
-# 
+#
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


jofemodo:
Hi @steveb !

For your next contribution, as you have write access on this repository, i think it's better you create your development branch on it. In such a way it's more easy to test changes on zynthian.

jofemodo:
After this change, controller areas don't work as "touch button". It seems your second change is too tight and it should be relaxed a little bit.

jofemodo:
In fact, your second change doesn't work at all, at less on the ZynScreen display, so i must reject this change. I will accept the first one: "If onscreen buttons are enabled, never use controllers as buttons"

Regards,

steveb:
> Hi @steveb !
> 
> For your next contribution, as you have write access on this repository, i think it's better you create your development branch on it. In such a way it's more easy to test changes on zynthian.

will do

steveb:
> In fact, your second change doesn't work at all, at less on the ZynScreen display, so i must reject this change. I will accept the first one: "If onscreen buttons are enabled, never use controllers as buttons"

OK, I'll take out the canvas_motion_set_value mechanism and restore canvas_motion_count and motion_rate

jofemodo:
I've already done it and it's already merged on testing.

Thanks!

PR #406, opened 2021-10-25
Zynmixer

riban-bw:
Request to merge zynmixer into testing.

Zynmixer is a library that provides a 16 stereo channel audio mixer. Each input has level (fader) adjustment, mute, solo, pan / balance and mono. Solo a channel will silence all other non-solo channels, i.e. solo channel 1 and hear channel 1 then also solo channel 2 and hear channel 1 and 2. The main mix bus has level (fader), mute, pan / balance and mono. All 16 stereo channels are mixed to a stereo output.

A control screen is implemented in TKInter which shows some / all of the 16 channels depending on screen resolution. (This may be changed in future to allow user configuration of quantity of channels displayed.) It allows control of each fader, mute, solo via hardware switches or touchscreen. Channel selection may be done by touch or using BACK encoder. Only the channels with a populated layer are displayed. If more channels are available the screen will horizontal scroll. Main fader is always shown and is accessed by scrolling far right or via touch. If _any_ channel is solo then the main fader strip shows SOLO. Main solo may be used to disable _all_ current channel solo. Bold press SELECT or press and hold the mute/solo button access individual channel control The fader for the channel and the main fader are shown on right and other control elements are available like balance (pan). Direct access to corresponding root layer is via this screen.

Currently the mixer is fixed at 16 channels, one per root layer (chain). This may change in the future, depending on system load. The mixer does not seem to cause significant load with 16 Fluidsynth layers playing Take 5 SMF.


riban-bw:
The main purpose of the PR is to add the mixer into the Zynthian chain. (Note that snapshots will need to be updated to correct the audio routing - maybe we can consider some automation.) The UI is a proof-of-concept (that works fine). I don't want UI discussions to stop merging the core mixer functionality. We can discuss and test different UI implementations once we have a core to interface with.

riban-bw:
@jofemodo asked if the fixed quantity of channels wasted CPU or Jack time. I just checked and channels are not processed unless there is audio routed to them, e.g. if only three channels are in use then only three are processing audio and the rest are silently ignored.

jofemodo:
Nice @riban-bw! I will test it carefully and merge at it's. I have a proposal for future improvement:

+ Let's include jackpeak's functionality in the mixer. We will reduce by one the number of jack clients and will save some RT CPU cycles.

Thanks Mr.!

riban-bw:
As if by magic... Peak programme meter for each channel and main bus are already implemented in mixer. Jackpeak is not used in zynmixer branch. I have been considering how we monitor the main output post-fx and thick we need to add an insert point in the output. I planned this last night but did not start to implement on the train, late at night after a long trip!

jofemodo:
Supernice! I just merged it. I must say i really like the implementation you did. The UI can (and must!) be improved, but it's a very good starting point.

Thanks a lot!

jofemodo:
I commited some adjustments for improving readibility on small screens.

PR #408, opened 2021-10-27
Fixes issues with refactor of menu to UI base class

riban-bw:
I had forgotten that I had started to consolidate the menu code into the UI base class but this was incomplete. I have now fixed the issues that were exposed by enabling touch widgets. Although this should have been submitted as a separate PR it will be significant work to separate the code so I humbly request this fix be merged rather than going back to square one - especially since zynmixer has been announced in the forum!

### DIFF
--- a/zyngui/zynthian_gui_base.py
+++ b/zyngui/zynthian_gui_base.py
@@ -2,13 +2,13 @@
 # -*- coding: utf-8 -*-
 #******************************************************************************
 # ZYNTHIAN PROJECT: Zynthian GUI
-# 
+#
 # Zynthian GUI Base Class: Status Bar + Basic layout & events
-# 
+#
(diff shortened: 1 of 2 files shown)


jofemodo:
OK! Let's go with it. Anyway, i would prefer to have all this stuff in a separated base class "zynthian_gui_xbase" that inherits from "zynthian_gui_base". Most of the "standard" screens from zynthian doesn't need all this stuff.

riban-bw:
zynthian_gui_xbase - we need to discuss what the zynthian gui should be built on. I felt making menus available universally and consistently to be advantageous if we start to use them more widely. There may be existing UI that would benefit but maybe we inherit touchscreen control from a separate class. Discussion for elsewhere though! Cheers!

PR #411, opened 2021-11-04
zynthian/zynthian-ui#351 Pianoteq PRO fixes

mheidt:
Fixing label issues with Pianoteq PRO


jofemodo:
I'm not fully confident of this change working on legacy 6.X. but i have not a standard or pro license, so i can't test.
Let's go forward and be prepared to fix it if somebody complains about it.

PR #417, opened 2022-02-10
Added new widget for displaying looper status.

looperlative:
New widget added and LP3 lv2 attached to the new widget.

### DIFF
--- a/zyngine/zynthian_engine_jalv.py
+++ b/zyngine/zynthian_engine_jalv.py
@@ -79,7 +79,8 @@ class zynthian_engine_jalv(zynthian_engine):
 	plugins_custom_gui = {
 		'http://gareus.org/oss/lv2/meters#spectr30mono': "/zynthian/zynthian-ui/zyngui/zynthian_widget_spectr30.py",
 		'http://gareus.org/oss/lv2/meters#spectr30stereo': "/zynthian/zynthian-ui/zyngui/zynthian_widget_spectr30.py",
-		'http://gareus.org/oss/lv2/tuna#one': "/zynthian/zynthian-ui/zyngui/zynthian_widget_tunaone.py"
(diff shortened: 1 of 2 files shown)


jofemodo:
Hi @looperlative !

How can i test? Where can i download the plugin?

Thanks!

looperlative:
Demo version of plugin available in the following tar file:

[zynthian-lp3-basic.tar.gz](https://www.dropbox.com/s/zsut666pt7e3m8c/zynthian-lp3-basic.tar.gz?dl=1)

S1 is Record/Overdub trigger
S2 is Play/Stop trigger
S3 is the Configurable Trigger
S4 is Erase All trigger

PDF included in the tar file.

looperlative:
Please note that the plugin has numerous bugs at the moment, but the basic record/overdub/play/functions work for demo of the UI.

jofemodo:
Hi @looperlative !

When you say S1-S4. do you mean that LP3 reacts to Program Change 1-4?
You shouldn't assume that S1-S4 are assigned to Program Change. This is only the default layout, but it can be easily changed from webconf. In fact, now we have the "Wiring Layout Profiles" and you can choose among different configurations. Also, you would like to use MIDI learning for using an external controller for triggering. 

For instance, I'm testing with the new Z2 prototype and i can't trigger anything because there is no S1-S4 assigned to Program Change.

Moreover, Program Change should be reserved for changing subsnapshots / presets. Imagine i have this audio chain:

   OBXd Synth => LP3 => Audio Output

and i've defined some subsnapshots (ZS3) that i want to select using Program Change. It's going to be problematic if LP3 use Program Change for triggers.

I would suggest you to use LV2 ports for triggering. CC-switches would work too and will allow that LP3 works with the S1-S4 buttons out-the-box when using the right "wiring layout profile" (aka, no need to MIDI learn for using S1-S4).

For the rest, very good work!! Congratulations!! I hope you want to share your code some day ;-)

Regards,

PR #420, opened 2022-02-20
Fixes #610 - high CPU when showing X42 Tuner GUI

riban-bw:
Marks tuna#one as broken ui.


jofemodo:
Agree! We don't need this GUI ;-)

PR #421, opened 2022-03-17
Bug/623 mixer control issues

riban-bw:
This refactors the mixer view to use the global layer list. It fixes issue 623 and also provides better user interface, e.g. users can now access all layers (chains) including MIDI only which was lacking in previous implementation. There is no direct relationship between mixer channel and chains as before (where we checked for audio routing). Such a link may be worth re-implementing in a considered way but this implementation is far more user friendly.

### DIFF
--- a/zyngui/zynthian_gui_mixer.py
+++ b/zyngui/zynthian_gui_mixer.py
@@ -794,6 +794,7 @@ class zynthian_gui_mixer(zynthian_gui_base.zynthian_gui_base):
 				self.visible_mixer_strips[offset].set_layer(None)
 			else:
 				self.visible_mixer_strips[offset].set_layer(layers[index])
+		self.main_mixbus_strip.redraw_controls(True)
 		
 
 	# Function to detect if a layer is audio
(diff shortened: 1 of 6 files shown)


zyngui/zynthian_gui_mixer.py
```diff
@@ -726,6 +726,13 @@ def __init__(self):
 		self.main_canvas.tag_bind("reset_balance_button", "<ButtonRelease-1>", self.on_reset_balance_release)
 		self.main_canvas.tag_bind("cancel_button", "<ButtonPress-1>", self.on_cancel_press)
 
+		# Horizontal scroll (via mouse wheel) area
+		legend_height = self.visible_mixer_strips[0].legend_height 
+		self.horiz_scroll_bg = self.main_canvas.create_rectangle(0, self.height - self.legend_height, self.width, self.height, width=0)
```
riban-bw: This blocks mouse events reaching the strips so blocks touch navigation to control view / layer options. Need to find a way to provide both event handlers.

jofemodo:
Tested and merged.  Good work!! ;-)

PR #422, opened 2022-03-23
Fixes #630 - Refresh mixer after program change

riban-bw:
Simple fix - remove check for mixer screen shown upon show. Hopefully it won't cause loops or flicker. Tests of common workflows seem okay.


jofemodo:
It seems to not cause problems.

PR #423, opened 2022-03-23
Use heart character to indicate favourites

riban-bw:
I thought it would be nice to show some love for our favourites so have submitted this PR which uses a heart character (❤) to indicate favourites rather than an asterisk (*). I have tested with all the fonts included with the standard Zynthian.


jofemodo:
I love it!

jofemodo:
See this: 602269f

riban-bw:
Sorry! Code from a development branch somehow made it in.

PR #424, opened 2022-03-24
Adds bold touch topbar to show main menu (by default)

riban-bw:
I have found mouse / touch navigation to be limited, getting stuck in screens and needing to access keyboard / encoders. This reduces this problem by providing a bold press action for the topbar and mapping it by default to show the _Main_ menu.


jofemodo:
Good idea!

PR #425, opened 2022-03-24
Use effect name (without JV/ prefix), indentation and curvey arrows t…

riban-bw:
…o show chain


jofemodo:
Great Brian! It improves quite a bit the readibility of the FX-chain

PR #427, opened 2022-03-25
zynthian.sh: Display exit code on error splash screen

polluxsynth:
When zynthian_gui.py exits with an unexpected exit code,
display it on the error splash screen to potentially aid
fault finding.

This has helped me diagnose the reason for zynthian_gui.py crashing.


jofemodo:
Thanks! I will improve the format a little bit ;-)

PR #428, opened 2022-03-28
Fix closing crash

polluxsynth:
It looks like the problem with Zynthian crashing when closing down was that the JACK client was still active, so deactivate it when shutting down the autoconnector.

### DIFF
--- a/zynautoconnect/zynthian_autoconnect.py
+++ b/zynautoconnect/zynthian_autoconnect.py
@@ -715,6 +715,7 @@ def start(rt=2):
 	# Start Autoconnect Thread
 	thread=Thread(target=autoconnect_thread, args=())
 	thread.daemon = True # thread dies with the program
+	thread.name = "autoconnect"
 	thread.start()
 
 
--- a/zynautoconnect/zynthian_autoconnect.py
+++ b/zynautoconnect/zynthian_autoconnect.py
@@ -725,6 +725,7 @@ def stop():
 	acquire_lock()
 	audio_disconnect_sysout()
 	release_lock()
+	jclient.deactivate()
 
 
 def is_running():
--- a/zyngine/zynthian_engine_mixer.py
+++ b/zyngine/zynthian_engine_mixer.py
@@ -415,6 +415,7 @@ class zynthian_engine_mixer(zynthian_engine):
 
 		self.sender_poll_enabled = True
 		thread = threading.Thread(target=runInThread, daemon=True)
+		thread.name = "engine mixer"
 		thread.start()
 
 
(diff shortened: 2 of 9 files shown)


jofemodo:
Ohhh! So simple, so nice. Congrats!
And thanks a lot!

I will adjust a little bit the thread names, of course ;-DD

jofemodo:
I will merge on testing without testing because the solution is so simple, that it can't harm.
I'm enjoying the end of the snow season, lost with my family in the top of a mountain, lodged in the f***g "The Shining" hotel, you can believe me!! ;-DD But i missed to take a zynthian with me .... OMG!!

polluxsynth:
Stuck in a big empty hotel without a Zynthian must be ... really hard ... even if you have your family with you ... :-)

Please do adjust the thread names, as you can see I don't really understand what each thread does, just gave it a name based on the surrounding code so I could see which threads were remaining at power off.

I think when you have the time to consider the fix also for the stable branch. I haven't verified this, but when I initially started up my Zynthian after building it, before the network was properly configured, I would sometimes get the Error screen when powering off, and I'm wondering if it was the same problem.

By the way after some more testing, the comment in the second commit in this PR that the fix also removes the Dummy thread is not always true. When the system is stopping and zynthian_guy.py prints out "Awaiting threads to end ...", I've seen three combinations (logically I would expect there to be a fourth one too):
1. 2 threads: MainThread and engine mixer
2. 3 threads: MainThread, engine mixer and control
3. 3 threads: MainThread, engine mixer and Dummy-N (where N is a fairly small number, like 3, 7 or 10) (I'm suspecting this number corresponds to the number of times the autoconnector is stopped/started while the application has been running, or simply the number of layers, but I haven't tested it further).

Whatever threads are there at the start of the loop, they remain all through the wait period, even if extended to 10 seconds (n=100). I'm wondering if this has to do with Pythons thread model, and during the wait period, the other threads never get a chance to run and terminate.

At any rate, in none of the cases does the application crash when exiting so the fix is still valid - I've not had a single power off crash since the fix was introduced. But I still thought I'd point out the bit about the Dummy thread still remaining sometimes.

jofemodo:
Don't waste too much time with the "dummy" thread. There are some "phantom" threads that we are not opening explicitly, like the ones opened by pexpect, but as far as they are not causing issues, we can ignore them ;-)

polluxsynth:
Agreed, I'm not worrying about it. But before the fix there was always a Dummy thread, and afterwards there mostly wasn't so I concluded that the Dummy thread was related to the xrun callback or perhaps the JACK client thing itself, and I wrote a note about it in the commit message, which apparently was not completely true so I just wanted to note that fact.

I have noted though that before the fix, if I changed n to 100 and added a printout for each time around the wait loop, I would get xrun callback printouts around the n=69 mark, whereas with the fix, there are no callback printouts at all during the loop, indicating that the fix actually does close the JACK client down, in particular so that no messages are received from it.

PR #429, opened 2022-03-29
Read get_throttled from sysfs

### LINKED ISSUE #412: XRUNs caused by vcgencmd
I've been investigating if the Raspberry Pi 4B (4GB) running Zynthian can manage a Jack period of 128 frames without xruns. My focus has been on the Salamander Grand Piano using LinuxSampler.

I'm using the InnoMaker HiFi DAC Audio Hat with jackd options `-P 70 -t 2000 -s -d alsa -d hw:1 -r 44100 -p 128 -n 2 -X raw` and RBPi Audio disabled.  With this configuration I experienced regular xruns.

I used [trace-cmd](https://www.trace-cmd.org/) on the Pi and [Kernel Shark](https://kernelshark.org/) offline to explore the cause of the xrun.  In the screenshot below you can see the regular pattern of real-time priority processes running each time Jack processes a period.  This pattern is interrupted around the cursor position. The interruption is caused by a fork system call, seen as the highlighted `copy_process` event. 

As far as I can tell, `fork` blocks all threads, even those with real-time priority.

xrun-p=128

Using [GDB](https://www.gnu.org/software/gdb/) I was able to identify the culprit as the fork system call required by the regular poll of `vcgencmd` to check for undervoltage and overtemp events. Once I commented out this section from `zynthian_gui.py` the xruns stopped:

```
try:
        # Get ARM flags
        res = check_output(("vcgencmd", "get_throttled")).decode('utf-8','ignore')
        thr = int(res[12:],16)
        if thr & 0x1:
                self.status_info['undervoltage'] = True
        elif thr & (0x4 | 0x2):
                self.status_info['overtemp'] = True

except Exception as e:
        logging.error(e)
```

In fact, I can now run the Salamander grand with a period of 64 frames with almost no xruns.

I'm sure there are a few ways we could retain this feature without triggering a fork.  One option would be to run an entirely separate process that regularly runs vcgencmd and puts the output in a file to be read by zynthian-ui.  

I do wonder though, if a better approach might be to pull the real-time priority jack processes out of zynthian-ui instead. This would make the performance of the zynthian-ui app less critical. 

Since I haven't spent a lot of time studying the zynthian software architecture I would appreciate others input before spending any time on even the simpler solution I've suggested.

steveb:
This avoids calling vcgencmd, which causes XRUNs in some environments.

Fixes #412

### DIFF
--- a/zynthian_gui.py
+++ b/zynthian_gui.py
@@ -39,7 +39,6 @@ from pathlib import Path
 from time import monotonic
 from os.path import isfile
 from datetime import datetime
-#from vcgencmd import Vcgencmd
 from threading  import Thread, Lock
 from subprocess import check_output
 from ctypes import c_float, c_double, CDLL,c_char_p
@@ -209,14 +208,13 @@ class zynthian_gui:
 		self.zynautoconnect_audio_flag = False
 		self.zynautoconnect_midi_flag = False
 
+		self.get_throttled_file = None
+
 		# Create Lock object to avoid concurrence problems
 		self.lock = Lock()
 
 		# Init LEDs
 		self.init_wsleds()
-
-		# Create vcgencmd instance
-		#self.vcgencmd = Vcgencmd()
 
 		# Load keyboard binding map
 		zynthian_gui_keybinding.getInstance().load()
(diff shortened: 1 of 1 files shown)


jofemodo:
Could be advantageous to keep open the sysfs interface? We would avoid extra "open" syscalls ...

steveb:
> Could be advantageous to keep open the sysfs interface? We would avoid extra "open" syscalls ...

Done

steveb:
Also I think this would deserve to be in stable, and I actually developed it against stable. So if you want a branch which merges cleanly to stable then check out https://github.com/steveb/zynthian-ui/commit/3b81cbb8e13b9d947f91a3927bad4c0546e36831

jofemodo:
Done! Also merged on stable.

Thanks a lot!

PR #436, opened 2022-04-07
Bug/635 touchscreen calibration

riban-bw:
Use coord matrix to support calibration of rotated display / touchscreen. Tested with all four rotations angles.


jofemodo:
Thanks @riban-bw !

PR #439, opened 2022-04-21
Bug/639 snapshot navigation

riban-bw:
Snapshot navigation fixed and improved.


jofemodo:
I will test ASAP

jofemodo:
Hi @riban-bw!

It's not working for me. First, it didn't loaded my last snapshot, instead it loaded empty,
When i try to explictly save the last snapshot, it doesn't work. In fact, i can save(overwrite) a snapshot. It simply does nothing.
In the other hand, saving as new does work OK.

For simplicity, i would remove the "create copy" option. It is redundant. You simply load a snapshot, modify and save as new. "Almost everybody" feel comfortable with this workflow and there is no need to have a redundant option for this. Also, it's not so trivial to implement. Less is more ;-)

Regards,

jofemodo:
Ahhh! Regarding "Set Program", numering is confusing. If my snapshot is "008-blablabla.zss" and try to change the program, it highlights program "07". 

First, MIDI program numbering must be coherent over all UI screens (and also webconf), so we should have a fixed criteria. IMHO, we should be transparent. No MIDI numbering translation anywhere in zynthian. Programs numbers start in 0. Continuous Controller numbers also does. Same for notes, etc.

**Rationale:** Translating numbers creates confusion in people that understand how MIDI numbering really works. You are punishing the "good boys" that have done their tasks. Showing the raw numbers creates confusion in people that doesn't understand how MIDI numbering works. You are teaching the miseducated. Also, translating numbers is and will be always a source of coding issues & bugs. We are avoiding a lot of stupid bugs by simply sticking to the MIDI spec as it is. 

My BIG doubt is with MIDI channels. My proposal is this, let it be as currently is and let's think of changing to 0-based midi channels when doing the chainification. Anyway, i have doubts. 

**Rationale:** Most of controllers i have use 1-based MIDI channels, but regarding the rest (program numbers, CCs, etc) , there is no fixed rule. It's hard to find a device using 0-based MIDI channels. Perhaps we will confuse to the "good boys" too and we don't want to do this, but in fact, it's not really coherent to do other way. What do you think?

Regarding this case (program numbers for snapshots):

+ AFAIK, the UI already uses 0-based program numbers when loading snapshots.

+ The webconf's snapshot manager also uses 0-based program numbers.

If you agree, i will change ZS3 screen numbering and i will locate other places to be changed, including webconf. I will not change MIDI channels yet.

Thanks!

riban-bw:
> It's not working for me
It is working for me. I wonder whether your subsequent messages mean you figured out what was going wrong?

> remove the "create copy" option
I will change the title of "Create Copy" to "Save as..." and not append "(copy)" to the file name shown in the OSK.
I will remove the "Save" option from the context menu.

> highlights program "07"
I see that behaviour and will take a look. MIDI program change has always been a bit ambiguous. I think we can use zero based for this. MIDI channel has never been ambiguous. It is always presented to the user as 1-16 but passed over the wire as 0-15. I agree this is inconsistent but it will make sense to maintain common practice. Defo keep channel as 1 based. Everything else seems to be zero based in user land.

I have a bit to do this morning but will work on this later so that hopefully when you wake up on Monday you will see the fixes I have done and work from that.

jofemodo:
No no, Save (overwriting) option still doesn't work for me.
Regarding  "create copy" ...

+ "Save as" already exist in the upper menu and it's OK like that.
+ "Save" is perfectly OK as it's in the snapshot context menu (when it works).
+ "Create Copy" should be removed because it's redundant. That's all. 

So the snapshot context menu would be like this:

+ Load
+ Load Layers
+ Load Sequences
+ Save
+ Rename
+ Set Program
+ Delete

Thanks!

jofemodo:
BTW, OK to keep the 1-based MIDI-channels and use 0-based numbers for the rest: Programs, CCs, etc.
I already changed ZS3 screen accordingly.

jofemodo:
Re-confirmed. Save options doesn't work.
Steps to reproduce:

+ Clean All
+ Create a Synth layer (ZynAddSubFX)
+ Open snapshot menu, choose an existing snapshot and click save.

**Observed:** No confirmation screen is shown. The snapshot is not overwriten. Mixer screen is shown.
**Expected:** Confirmation screen is shown.The snapshot is overwriten if "yes" is selected. Mixer screen is shown??

riban-bw:
I have fixed Save snapshot and changed the context menu. Also hide some irrelevant / dangerous options for "Last State".

jofemodo:
You are fast my friend!! I was starting to think i'm totally mad when testing in a second zynthian and everything was working. Then i saw your latest commit ... jejeje!

Let me test it carefully and i will merge it ASAP.

Thanks!

jofemodo:
BTW, could you test the masterfx branch? I would like to merge everything and start generating testing SD images.

jofemodo:
Snapshot workflow now feels a lot better.

riban-bw:
Even better now! I see you merged and deleted but I fixed a couple of bugs. I needed to complete this before moving on. Will look at masterfx tomorrow.

jofemodo:
I see the improvements on last snapshot submenu. I like it ;-)

Please, remember i created some show_screen() shortcut functions for better readibility: 

+ show modal() (it already existed before)
+ show_screen_reset()
+ replace_screen()

Regards!

PR #441, opened 2022-04-25
Feature/614 preset management

jofemodo:


zyngui/zynthian_gui_bank.py
```diff
@@ -71,7 +71,8 @@ def select_action(self, i, t='S'):
 		self.zyngui.show_screen('preset')
 
 		# If there is only one preset, jump to instrument control
-		if len(self.zyngui.curlayer.preset_list)<=1 and self.zyngui.curlayer.preset_list[0][0]=="":
+		#TODO: This is inconsistent behaviour which may confuse users
```
jofemodo: It's not inconsistent, it was wrongly commented ;-D
It should be "If the only preset is the empty preset (aka, there is no presets), jump to instrument control"

riban-bw: I have tweaked the comment in my next commit. Also the code is prone to exception. If len == 0 then accessing [0] will error. I have changed code to be safe in next commit.

PR #443, opened 2022-05-04
Feature/audio player

riban-bw:
Adds bespoke audio player as a Zynthian engine. This allows playback of wav, ogg, etc. files through effects chains to a mixer strip. Files are presented as presets and folders as banks.


riban-bw:
This will resolve zynthian/zynthian-issue-tracking#645. There are a few things yet to do that I may need some help with, e.g. setting the controls from the library / engine.

zyngui/zynthian_gui_control.py
```diff
@@ -330,21 +330,79 @@ def back_action(self):
...
-			i = 0
-		self.select(i)
-		self.click_listbox()
+		if self.controllers_lock and self.mode=='control' and self.zcontrollers:
+			try:
+				self.zcontrollers[3].nudge(1)
+			except:
+				pass
+		else:
+			super().select_up()
```
jofemodo, line 454: No my friend!!! You can't change the behaviour of things you don't understand like this!! Are you missing we have a new prototype called Z2? I will revert this change. Please, don't change things like this. I will send you a Z2 ASAP so you can test with it. Until then, please, think wise. There are lots of improvements & adjustments specifically done for the Z2.

jofemodo: Also, this is very VX-centric. We are moving to something more flexible. The classic "4 rotaries + 4 switches" still have to work, but it shouldn't be hardcoded anymore. I mean, we shouldn't hardcode switch/knobs actions like this.  Not in new code. In fact, we should refact old code to remove this.

jofemodo:
Controller refact "side effects" (part I)

**setbfree**
    + "Rotary Speed" controller doesn't work like expected. Can't change to rotary fast and "off" is "slow", etc.
    + "DrawBars" are inverted, When you move up, they go down. When you move down, you go up.

jofemodo:
BTW, i just noted that AudioPlay CUIAs doesn't work anymore. This also breaks quite a bit the Z2 behaviour.
```
		#TODO: Handle CUIA audio playback
		#elif cuia == "START_AUDIO_PLAY":
		#elif cuia == "STOP_AUDIO_PLAY":
		#elif cuia == "TOGGLE_AUDIO_PLAY":
```
It's not a big problem, but i would need to know if you already have plans for restoring this and how do you intend to do it.

zyngui/zynthian_gui_control.py
riban-bw: I am aware that these changes are likely to break stuff and have been methodically working through the modules to fix errors. Of course it is challenging to test the Z2 stuff but I had intended to look at its behaviour but tried to make the initial changes hardware agnostic.

VX-centric? Not sure what that means.

riban-bw:
Playback is per instrument and can be triggered with CC. It may be part of the API to trigger such actions similar to CUIA. The old CUIA for audio player was not obvious which audio it would play. The new player may have many instances, each with own CC to control via MIDI learn for example. I suggest we don't implement CUIA for audio player.

I am methodically working through functional tests. I have not yet reached setBfree (beyond checking it loads and plays).

I didn't realise / had forgotten there was a PR for this development branch hence have been working heavily on it.

jofemodo:
I think you need a Z2 urgently ...  ;-)
I hope to send you one very soon, but they are difficult to get, mate!!

jofemodo:
**Selector Separators:**
It seems to be some problem with "skip_separators" on zynthian_gui_selector. Try going slowly throw a separator, it start to jump crazy up & down. This is due to the recent refact for sure ;-)

jofemodo:
**Navigating Controller Screens** with the selector knob doesn't preview the controllers pages like before.

jofemodo:
> It is not possible to select mixer chain with QWERTY keyboard. Previously this was done with the up/down arrow keys emulating the select encoder.

You should use the new ARROW_LEFT/RIGHT CUIAs.

jofemodo:
Navigating Controller pages & skip_separators seems to be solved with my latest commit.

riban-bw:
I think this is probably good enough to merge. Issues with encoder control, etc. should be tracked and resolved under separate tickets / dev branches. I will report such issues in github issue tracker.

PR #446, opened 2022-06-02
Feature/internet radio

riban-bw:
A little light relief... Some of our community quite like the ability to listen to the radio through their Zynthians. It can be a feed during rehearsal or recording sessions to amuse. It can be a feed to a live audience. It can be background listening whilst the Zynthian is idle. It can be a random source of audio to use artistically.


riban-bw:
This is now updated and merged to chain_manager so closing the PR.

PR #449, opened 2022-07-01
Fix ZS3, snapshot save for layers with no midi channel

steveb:
Not all layers have MIDI channels (mod-ui) but the following features are
making assumptions that the layer midi channel number will not be null:
- Snapshot save
- ZS3 program change learn
- ZS3 program change recall

This change fixes these features for the mod-ui engine by allowing for
None midi channel numbers.

This is a stable-only change to fix broken features for the mod-ui engine. ZS3 has been significantly reworked in testing and likely doesn't have these issues.


jofemodo:
Thanks @steveb! It's merged now.

jofemodo:
BTW, next stable is really close now ;-)

PR #453, opened 2022-11-06
Pianoteq rpcjson api

riban-bw:
This PR provides integration of Pianoteq 7.5 and later via RPC-JSON API. It adds a new class for this integration and Zynthian choses the class based on the version of the Pianoteq executable.


zyngine/zynthian_controller.py
```diff
@@ -137,13 +137,15 @@ def _configure(self):
...
+				value_range = self.value_max - self.value_min
+				if n == 1:
+					self.ticks.append(self.value_min)
+				elif self.is_integer:
 					for i in range(n):
-						self.ticks.append(self.value_min + int(i * value_range / n))
+						self.ticks.append(self.value_min + int(i * value_range / (n - 1)))
 				else:
 					for i in range(n):
-						self.ticks.append(self.value_min + i * value_range / n)
```
jofemodo, line 146: This should be done in a separate PR. It affects all engines!

zyngui/zynthian_gui_save_preset.py
```diff
@@ -38,7 +38,7 @@ def save_preset(self):
 			self.save_preset_create_bank_name = None
 			self.save_preset_bank_info = None
 			self.layer.load_bank_list()
-			if not self.layer.bank_list or not self.layer.bank_list[0][0]:
```
jofemodo, line 41: Same for this. it's a new feature, not related to Pianoteq API implementation, and it affects all engines.

zyngine/zynthian_engine_pianoteq_jsonrpc.py
```diff
@@ -0,0 +1,696 @@
...
+	PIANOTEQ_JACK_PORT_NAME = PIANOTEQ_NAME
+
+if PIANOTEQ_PRODUCT != "STAGE":
+	PIANOTEQ_CONFIG_FILENAME = "{}.prefs".format(PIANOTEQ_NAME)
+else:
+	PIANOTEQ_CONFIG_FILENAME = "{} {}.prefs".format(PIANOTEQ_NAME, PIANOTEQ_PRODUCT)
+
+PIANOTEQ_CONFIG_FILE = PIANOTEQ_CONFIG_DIR + "/" + PIANOTEQ_CONFIG_FILENAME
+
+
```
jofemodo: Could/should we split these auxiliary functions so we have a single/common copy of them for both implementations?

zyngine/zynthian_engine_pianoteq_jsonrpc.py
```diff
@@ -0,0 +1,696 @@
...
+		return False
+		
+
+	@classmethod
+	def zynapi_remove_preset(cls, preset_path):
+		if preset_path[-4:].lower() != ".fxp":
+			return False
+		os.system(f"rm '{preset_path}'")
+		return True
+
```
jofemodo: Perhaps having a "zynthian_engine_pianoteq_base" class have sense, specially if we plan to keep compatibility with pianoteq 6 in the future. I know it's an extra effort, but code will be easier to maintain as we avoid duplication. Also, it's a good exercise and it normally helps to improve code quality. Pianoteq is a key engine for the project so the effort is worth it.

zyngine/zynthian_controller.py
riban-bw: Possibly but this line of code is not used by any engine until the new pianoteq implementation invokes it. I loaded every type of engine in testing branch and none trigger this line.

zyngui/zynthian_gui_save_preset.py
riban-bw: Possibly but the enhancement is added to facilitate functionality in pianoteq and is not previously implemented.

zyngine/zynthian_engine_pianoteq_jsonrpc.py
riban-bw: Yes we should! I will look at this.

zyngine/zynthian_controller.py
riban-bw: Actually, the change before this one is more impactful and needs more consideration.

riban-bw: I have submitted https://github.com/zynthian/zynthian-ui/pull/459

zyngine/zynthian_engine_pianoteq_jsonrpc.py
riban-bw: We could use the new pianoteq class and derive a class from it that implements legacy (<7.5) control.

linuxnow:
I have tested the changes and they work with Pianoteq 7.5 Stage. I will use them on real stage performance :)

zyngine/zynthian_controller.py
riban-bw: I have created individual PRs for each change that has wider impact. They are also consolidated into this PR.

zyngine/zynthian_engine_pianoteq_jsonrpc.py
riban-bw: I have moved all the common code to "pianoteq.py". Some functions are static and called from "pianoteq6.py". I decided not to subclass Pianoteq as it proved too challenging. It was also too challenging to consolidate to a single class, hence there are still two classes: pianoteq and pianteq6. The latter is for pianoteq versions < 7.5.

zyngine/zynthian_engine_pianoteq_jsonrpc.py
riban-bw: See comment above in other conversation. There remain 2 non-inherited classes.

riban-bw:
@linuxnow there have been more (quite significant) changes. May I ask you to retest?
@jofemodo I believe this is now fully working.

linuxnow:
I have installed the consolidated version, with testing already merged. And everything works fine. I'm going to use it on stage today, so I'm pretty sure :)

I've tested only standard presets; now that I can modify all the params I don't need top create a preset fomr the original interface. I will test them on monday.

linuxnow:
I have some feedback, a small thing to fix.

The sustain pedal, which I presume is always in the same CC number,  doesn't work by default. Instead the expression pedal is the one that uses the CC number.

I had to unplug the expression pedal to midi learn the sustain pedal in all the ZS3 subsnapshots, one by one.

I'm completely sure that I used the correct input as the rest of engines worked as expected.

riban-bw:
@linuxnow I do not see this behaviour. By default Piantoeq assigns MIDI CC 64 to sustain pedal, 66 to sostenuto pedal, 67 to soft pedal, 69 to harmonic pedal. I think this was the same with the legacy version. All of these are the MIDI standard assignment except 69 which MIDI suggests being "Hold 2" (whatever that might be).

Testing here, the sustain pedal works with a new instance of Pianoteq and one loaded from a legacy snapshot. You can see the MIDI CC number that the control has at the bottom of the controller, e.g. Sustain Pedal is on "Pedals#1" page of the control view.

linuxnow:
I have tried from scratch and I can corroborate that MIDI CC 64 is assigned to the Sustain Pedal as you say, as well as the rest of pedals.

I don't know where this CC was lost in my snapshots but it was not there until I relearnt it for each one. Anyway, it works as it should.

For the rest it has worked flawlessly! Great work.

linuxnow:
One smal bug: the Pianoteq webconf page does not work. In order to upload a new ptq I need to choose the testing version.

linuxnow:
Pianoteq 8 works out of the box (in testing branch too). But I have to reset the settings, something is not working well loading the Z23 stored settings.

linuxnow:
> One smal bug: the Pianoteq webconf page does not work. In order to upload a new ptq I need to choose the testing version.

This affects zynthian_webconf. I include the error:

Nov 19 17:02:11 zynthian2 zynthian_webconf.sh[427]: ERROR:web.log_exception: Uncaught exception GET /sw-pianoteq (192.168.1.3)
....
Nov 19 17:02:11 zynthian2 zynthian_webconf.sh[427]:   File "/home/pi/zynthian-webconf/lib/pianoteq_handler.py", line 60, in get 
Nov 19 17:02:11 zynthian2 zynthian_webconf.sh[427]: NameError: name 'PIANOTEQ_TRIAL' is not defined  
Nov 19 17:02:11 zynthian2 zynthian_webconf.sh[427]: ERROR:web.log_request: 500 GET /sw-pianoteq (192.168.1.3) 89.67ms


In zynthian_envars-sh the env var names seem correct:
export PIANOTEQ_TRIAL="0" 
export PIANOTEQ_PRODUCT="PRO"
export PIANOTEQ_VERSION="8.0.0"

linuxnow:
I have discovered the pianoteq_api branch in webconf, have merged it into testing and now the webconf works.

There's still something that is not right:: I have upgraded to the Pro release and, although the file is properly installed, I still see Stage instead of Pro and I cannot input the new license key. I have to input it using the VNC interface.

linuxnow:
With the new patch I miss the Instrument name in the Mixer:
-I see Pianoteq, just beneath Studio but thre's no Steinberg D HB.
-I see Pianote Basic buit not Electra
.MKI and MKII are shown beacuse it's part of the preset name

In the config page everything is fine.

linuxnow (changes requested):
I've been checking all the preset names in Pianoteq 8 and we could directly use them. In all the cases you get the bank name information and the preset information, without exception.

We can simplify the get_display_name function to directly return preset_name and deal with some special case that I cannot see,  or simply get rid of it.

linuxnow:
> There's still something that is not right:: I have upgraded to the Pro release and, although the file is properly installed, I still see Stage instead of Pro and I cannot input the new license key. I have to input it using the VNC interface.

Pianoteq 8 stores the preferences in Pianoteq80.prefs.
This might be the reason why Product and License Key are not right, they seem to come from the other config files ([Pianoteq75 STAGE.prefs or Pianoteq.prefs).

riban-bw:
I have previously made a change to preference file handling, using a specified file taker than calculating the default filename. This should work for PT8 but I haven't tested. I am currently traveling so unable to look for a few days.

linuxnow:
Everything works, I've used it on stage a few times already, but there are two small issues pending:
-there's no way to upload the Pianoteq 8 software and be asked the  new license key, you have to do it via VNC
-the patch that simplifies names eliminates useful information from the mixer and config screens ( commit aa7729a787023d8decc2f2ccb9bd547f17debb39).
 All preset names in Pianoteq have the necessary bank information, we can simplify or remove the whole  patch just using  the preset name. I don't know if it has other side effects as I'm not proficient in the code, otherwise I'd send a pull request.

But in general this branch could be merge as it improves the user experience by a large amount.

Thanks for the good work @riban-bw !

riban-bw:
Zynthian shows instrument/bank/preset scrolling in the topbar in the control view. The patch reduced the redundant duplication, i.e. where the bank or instrument name also appears in the preset name. There are some patches where this doesn't happen and in previous implementation (<7.5) some presets were not available. Please describe the information missing from mixer view. (We may change this soon to allow chain to be named as the chain can have several plugins, each with their own presets but maybe the instrument preset should take precedence.)

Is it only upload of 8.0 that fails? Can earlier versions still be uploaded?

linuxnow:
The information displayed in the control view is fine, as you say it skips unnecessary repetitions.
But in the mixer view you can only see Pianoteq and a too shortened version of the preset name
-instead of HB Steinway D Classical, which is the preset name, you can only see Classical.
-instead of Electra-Piano Basic you can only see Basic.

Maybe we could use the original preset name in the mixer view and the shortened name in the control view :)

I'll try to remove the preferences files and reinstall different Pianoteq versions and come back to you,

linuxnow:
Wow, merged! I think that the change is so big that it deserves a new stable sub-.release :)
Thanks for the great work!

linuxnow:
> The information displayed in the control view is fine, as you say it skips unnecessary repetitions. But in the mixer view you can only see Pianoteq and a too shortened version of the preset name -instead of HB Steinway D Classical, which is the preset name, you can only see Classical. -instead of Electra-Piano Basic you can only see Basic.
> 
> Maybe we could use the original preset name in the mixer view and the shortened name in the control view :)
> 
> I'll try to remove the preferences files and reinstall different Pianoteq versions and come back to you,

This still applies, the nformation in the mixer screen is not complete, we should only use the preset name than has everythin needed

riban-bw:
> Wow, merged! I think that the change is so big that it deserves a new stable sub-.release :) Thanks for the great work!

We are working on a substantial change to the core code which I hope will be ready within the next month. That will be a good time to make a new stable release.

>the information in the mixer screen is not complete
We are considering what is displayed in the mixer. We currently pick out what we think is the most important info, e.g. synth preset name is highest priority, but we may want to give users the ability to name chains in which case this may be appropriate. With a chain of processors it may not be easy to automatically decide what info is most useful to a user - this needs some thought. We can look at how best to extract the useful info from Pianoteq.

linuxnow:
> We are working on a substantial change to the core code which I hope will be ready within the next month. That will be a good time to make a new stable release.

I've seen the chain_manager branch, completely out of my reach, I don't understand how MIDO works yet.

> We are considering what is displayed in the mixer.

With simple engines it could be a class method that you can override per engine, this way you could just change it when needed.

With complex chains probably the best way is a user-defined name,.

If you need any testing just let me know.

PR #454, opened 2022-11-06
Fix zynthian/zynthian-issue-tracking#817: Fix preset restore after preload,

jofemodo:
so it's always done after exiting preset/bank screen.

### DIFF
--- a/zynthian_gui.py
+++ b/zynthian_gui.py
@@ -721,6 +721,9 @@ class zynthian_gui:
 			else:
 				return
 
+		if screen not in ("bank", "preset", "option"):
+			self.screens['layer'].restore_presets()
+
 		self.screens[screen].build_view()
 		self.hide_screens(exclude=screen)
 		if hmode == zynthian_gui.SCREEN_HMODE_ADD:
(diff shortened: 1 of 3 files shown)


zynthian_gui.py
```diff
@@ -721,6 +721,9 @@ def show_screen(self, screen=None, hmode=SCREEN_HMODE_ADD):
 			else:
 				return
 
+		if screen not in ("bank", "preset", "option"):
```
riban-bw, line 724: This function gets called a lot - one almost every screen navigation.

riban-bw: If pre-selection is toggled as favourite then this still fails, i.e. preset is restored.

riban-bw (commented):
I don't think this is sufficiently urgent to warrant a hot fix. It only affects Z2. Can it wait until a point release?

jofemodo:
It affects V4 too. For instance, from preset preloading, if you press bold-back to open main menu, engine-selected preset is not restored either, etc. 
Also, i think this could solve https://github.com/zynthian/zynthian-issue-tracking/issues/764. I "kind of reproduced" #764 when trying to solve this, and after the fix, i didn't see #764 again.
We should merge to testing.

PR #455, opened 2022-11-07
+ Fix zynthian/zynthian-issue-tracking#816 => Calculate root layers before calling zynautoconnect

### LINKED ISSUE #818: After receiving ZS3 PC, control screen is not refreshed.
**Describe the bug**
After receiving ZS3 PC, control screen is not refreshed. New active chain is not shown!

**Expected behavior**
Control screen  show the active chain always.

**Setup:**
 - Latest Stable 2211
 - Z2 & V4

**Additional context**
This is specially important in stage mode. In multitimbral mode, we could choose to "not restore" the active chain from the ZS3.

jofemodo:
so "drop PC" flag is correctly set for every chain.

+ Fix zynthian/zynthian-issue-tracking#818 => Refresh control screen after ZS3 program change. Remove GUI calls from core.

### DIFF
--- a/zynautoconnect/zynthian_autoconnect.py
+++ b/zynautoconnect/zynthian_autoconnect.py
@@ -39,9 +39,9 @@ from zyngui import zynthian_gui_config
 # Configure logging
 #-------------------------------------------------------------------------------
 
-log_level=int(os.environ.get('ZYNTHIAN_LOG_LEVEL',logging.WARNING))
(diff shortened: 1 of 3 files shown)


riban-bw (approved):
Looks fine. There seems to be some refactoring which will facilitate separating ui/data model, e.g. partial removal of zynthian_gui_layer::fill_list.

jofemodo:
Yes, i couldn't avoid to improve things after re-reading the code. Jeje!

PR #459, opened 2022-11-08
Fixes zynthian/zynthian-issue-tracking#821

riban-bw:
Corrects error calculating values for discrete controls from labels.


jofemodo:
Steps to reproduce the error?
Controller class is a complex beast and it's used in many different ways depending of engines. It's really easy to fix something for one engine and break it for another one. This kind of changes are not trivial at all ...

PR #461, opened 2022-11-11
Prefer preset UID to name when loading snapshot

riban-bw:
This fixes some snapshot loading compatibility issues.
Prefer preset UID if available else use old behaviour (name).


jofemodo:
Open eyes with this change!!!

PR #462, opened 2022-11-27
Update zynthian_engine_fluidsynth.py

samriccijr:
Swap controllers 72 and 73 to match GM specification.


jofemodo:
Thanks @samriccijr !

PR #465, opened 2023-01-14
Hotfix/pianoteq blüthner

riban-bw:
This fixes Blüthner in Pianoteq. Only a stable hot-fix because the mechanism for this has changed in testing branch.


riban-bw:
Sorry - PR against wrong branch.

PR #469, opened 2023-01-29
Fix name of mod-ui pedal_snapshot command

steveb:
The current version of MOD-UI issues a pedal_snapshot command when the MOD-UI snapshot (zynthian preset) has finished loading.

Without this change all event processing stalls for 10 seconds while the engine waits for the pedal_preset command which never arrives.

### DIFF
--- a/zyngine/zynthian_engine_modui.py
+++ b/zyngine/zynthian_engine_modui.py
@@ -403,7 +403,7 @@ class zynthian_engine_modui(zynthian_engine):
 				elif command == "preset":
 					self.preset_cb(args[1],args[2])
 
-				elif command == "pedal_preset":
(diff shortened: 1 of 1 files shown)


steveb:
I'd really like this change in stable :)

steveb:
I've added another fix to this PR:

 modui: fix off-by-one in preset selection

The current approach in pedal_preset_cb results in an off-by-one issue
where layer.set_preset displays the preset one before the desired one.

This change switches both preset_cb and pedal_preset_cb to
set_preset_by_id instead of set_preset, which sets the correct preset.

jofemodo:
Thanks @steveb !
I will try to merge on stable ASAP.

jofemodo:
Done!

PR #470, opened 2023-02-04
modui: Use ~/.pedalboards/ for the bundlepath.

steveb:
~/.pedalboards/ is a symlink to
/zynthian/zynthian-my-data/presets/mod-ui/pedalboards, but the mod-ui save method has conditionals[1] which force saving as a unique name unless the bundlepath matches the configured path. This makes it hard to make changes to existing zynthian snapshots since the pedalboard name is always changing.

This change switches the path in the engine to the path mod-ui is configured for. Alternately we could carry a mod-ui patch to change the default[2] but that might be harder to maintain.

[1] https://github.com/zynthian/mod-ui/blob/master/mod/host.py#L2264
[2] https://github.com/zynthian/mod-ui/blob/master/mod/settings.py#L55

### DIFF
--- a/zyngine/zynthian_engine_modui.py
+++ b/zyngine/zynthian_engine_modui.py
@@ -85,6 +85,10 @@ class zynthian_engine_modui(zynthian_engine):
 		self.start()
 
 
+	def __del__(self):
+		pass
+
+
 	def reset(self):
 		super().reset()
 		self.graph = {}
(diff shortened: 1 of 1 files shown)


steveb:
Also:
 modui: override __del__ to do nothing

__del__ is called during garbage collection, which may be some time
after an engine is stopped, and the default implementation of __del__
also calls self.stop().

This causes a problem in the following scenario:
- Running a mod-ui already
- Navigate to Main -> Clean All
- Run mod-ui again from a snapshot
- At some point (when garbage collection is triggered), mod-ui is stopped

I really think __del__ should not stop by default, due to the
unpredictability of when garbage collection gets triggered on an
already stopped engine. But this change just fixes that for the mod-ui
engine.

jofemodo:
OK! You convinced to me. I will remove the "__del__" override in the "zynthian_engine" parent class.

Thanks @steveb !!

PR #474, opened 2023-04-29
ZynSeq fixes

riban-bw:
Fix missing colours in arranger.
Use queue for zynpad callbacks to avoid re-entrant crashes.
Fix libseq bank size issue.

These are important fixes to the sequencer.


riban-bw:
There is still a possibility of seg fault when changing bank/scene rapidly. This is an existing issue that I thought I had fixed in this patch. I need to investigate further so will cancel this PR until this issue is resolved.

PR #477, opened 2023-05-14
Fix logic for enable_touch_controller_switches

steveb:
When there are visible screen buttons, the controller touch interface doesn't need to be overloaded with button press actions. Accidental button presses are common when making small touch value changes, so if there is buttons on the actual screen the controllers shouldn't handle touches as buttons.

This seems to be the intent of enable_touch_controller_switches but the logic is inverted and also includes a Z2 wiring layout which may no longer be relevant.

This change also enables touching the top bar for back button presses, regardless of the value of enable_touch_controller_switches. This is because there is often no on screen alternative to trigger back button presses.

### DIFF
--- a/zyngui/zynthian_gui_base.py
+++ b/zyngui/zynthian_gui_base.py
@@ -325,14 +325,12 @@ class zynthian_gui_base(tkinter.Frame):
 
 	# Default topbar short touch action
 	def topbar_touch_action(self):
-		if zynthian_gui_config.enable_touch_controller_switches:
-			self.zyngui.zynswitch_defered('S', 1)
+		self.zyngui.zynswitch_defered('S', 1)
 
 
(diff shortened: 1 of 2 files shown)


jofemodo:
The Z2 wiring layout is very relevant. The logic was: 

> we don't want "touch switches" when using a Z2 device and on-screen buttons are not enabled

`if wiring_layout.startswith("Z2") and not enable_onscreen_buttons`

But we could change to something better, like this:

```
if check_wiring_layout("Z2", "V5") or enable_onscreen_buttons:
	enable_touch_controller_switches = 0
else:
	enable_touch_controller_switches = 1
```

> we don't want "touch switches" when using a Z2 or V5 device nor when on-screen buttons are enabled

Regards,

PR #481, opened 2023-11-02
Fix "allow_rbpi_headphones" to handle both V5 and older zynthians

multim0:
https://github.com/zynthian/zynthian-issue-tracking/issues/914


jofemodo:
Hi @multim0 !

I merged but fixed just after because there was a bug:

You missed the "self." in this line:

`if is_headphone_amp_interface_available():`

Please, test before PR ;-)

Thanks a lot!

PR #486, opened 2024-01-15
Fix zynthian/zynthian-issue-tracking#931

oscaracena:
Rename symbol `enable_audio_thru` to current param name `audio_thru`.


jofemodo:
Thanks Oscar!

PR #489, opened 2024-01-15
Fix zynthian/zynthian-issue-tracking#937

oscaracena:
`getStateChange/getProgress` already omits the endpoint. Passing `num_seq - 1` iterates from 0 to `num_seq - 2`, ignoring the last sequence number.


jofemodo:
This perfectly fix the issue!! Thanks a lot, Oscar!

PR #492, opened 2024-02-16
closes zynthian/zynthian-issue-tracking#951: chain_manager/get_chain_…

oscaracena:
…by_position() return chain object always


riban-bw:
Thanks - good spot!

PR #498, opened 2024-05-27
ctrldev: add offset parameter to addNote calls

oscaracena:


oscaracena:
@jofemodo I updated the comment, instead of replying with a new one, and thus may have not seen this PR. In that case, here you are 😉 If you saw it, but "no te da la vida", I understand perfectly, just ignore this! 😄

PR #501, opened 2024-07-12
V5touch: add V5-like touchscreen keypad to the UI

wanthalf:
Touchscreen V5 keypad implementation as discussed in https://discourse.zynthian.org/t/diy-upgrading-to-v5-or-the-question-of-new-buttons/9707
(NOT everything was tested in depth yet.)

Requirements and options:

- `python3-cairosvg` to support SVG icons (should fall back to PNG if not available)
- environment variable `ZYNTHIAN_TOUCH_KEYPAD` set to `V5` to activate this extension (without this setting, the UI should behave exactly as before)
- optional environment variable `ZYNTHIAN_TOUCH_KEYPAD_SIDE_LEFT` to activate  the left side layout of the keypad (TRUE by default)
- optional environment variables for setting custom onscreen buttons foreground colours (see `zynthian_wsleds_v5touch.py`)


jofemodo:
Hi @wanthalf !

Excuse the radio silence. I'm quite busy on summer and had little time to read your code and test the changes. I will do ASAP.

Regards,

wanthalf:
No problem. I see the tons of problems that must be finished and fixed ASAP! :)

wanthalf:
A question: once I noticed a state (that I cannot reproduce nor find in the code) where the (on-screen) wsleds were all turned off except the one indicating the active screen, which was pulsing in green. If there is such a state, I would suggest to change the "wscolor_off" from black to grey, so that the labels remain partially visible on the black background even when "turned off". The physical buttons also have their labels to some degree visible when the LEDs are turned off.

riban-bw:
That sounds like power save mode. In power save,  the screen is switched off. I guess you may be testing via VNC, not to observe the lack of screen output?

wanthalf:
I see I also overlooked a few more geometry settings in `zynthian_gui_control_xy.py`(never actually seen that screen). Also, many sizes are still computed from the full screen dimensions, not the reduced one in `zynthian_gui_config.py`  (e.g. size of the font, logo image, top bar etc.). Not sure how to best deal with the reduced dimensions. Maybe it would be better to  have two separate values for geometry such as `zynthian_gui_config.display_width` and `zynthian_gui_config.full_display_width`? (And not just subtracting the keypad width and height from  the full display size in (nearly) every screen object - as I am doing now?)

The only screens really using the full display dimensions are probably the splash and the touchscreen calibration (and `multitouch.py`, of  course)?

wanthalf:
No feedback yet, so I tried to make the integration a bit cleaner myself by moving it into `zynthian_gui_config` and  creating new attributes `screen_width` and `screen_height`, independent of the original `display_width`  and `display_height`.  (Hopefully also fixing the missing changes in `zynthian_gui_control_xy.py`.)

zyngine/zynthian_state_manager.py
```diff
@@ -938,6 +938,14 @@ def get_state(self):
         b64_data = base64.b64encode(binary_riff_data)
         state['zynseq_riff_b64'] = b64_data.decode('utf-8')
 
+        # TouchKeypad config
```
riban-bw, line 941: Any state/snapshot structure changes must be documented in zynthian_state_schema.py.

wanthalf: Thanks for hint!

PR #504, opened 2024-08-17
APC Key25: fix problem with pattern recording

oscaracena:


jofemodo:
Hi Oscar!
I've fixed the issue in the base code. This fix is not needed.
Thanks!!

PR #508, opened 2024-12-26
Fixes issues 1164 & 1168: switching to a new pattern was wrong; invalid LEDs status on mute/solo

oscaracena:


oscaracena:
@jofemodo @riban-bw I'm having this PR for a while, and as I want to keep it updated with vangelis (to avoid conflicts), it is growing... 😅 Is there any problem for merging it? It only affected my driver, so it should not present any conflict. If, as a result of my own merges on this branch, now the PR is difficult to review, I can make it again, starting with the latest vangelis branch.

Thanks in advance!

jofemodo:
No problem with the driver changes, but we need to be sure about the git management changes
Why not separating into 2 PRs? It would be easier to get the driver changes merged right now and we could verify carefully the git related changes.

oscaracena:
Sure, sorry for the noise. I'll make a new PR with only the right things :)

PR #512, opened 2025-01-17
Network MIDI 2.0: set port alias using Jack property UMPEndpointName

oscaracena:
This allows creating control drivers for devices connected using Network MIDI 2.0.


jofemodo:
Merged! Thanks @oscaracena !

PR #518, opened 2025-04-14
Fix copy pattern selection

xjaner:
This pull request fixes a problem I encountered when copying patterns: if you want to copy a pattern to another one whose number is close to it, it is not possible without this fix.

The cause of the problem is that the range of values for pattern number is so huge (from 1 to zynseq.SEQ_MAX_PATTERNS=64872) that when turning knobs # 3 and # 4 to define the destination pattern, you can only add 100 or 1000 to the current pattern, so it's impossible to add only 1 or 2 to the current pattern number.

With this fix I'm setting the `nudge_factor` to 1, so knob # 4 let's you add 1 to the current pattern and knob # 3 adds 100, making it possible to select any pattern number.

PD: A few months ago I reported another issue when copying patterns, I have the feeling I'm the only one using this feature! 😂


jofemodo:
Hi @xjaner ! I just merged your fix. 

FYI, the "copy pattern" functionality has been improved in our development branch and now it's more a "copy/paste" one. Perhaps it's not so "agile", but it's more clear and flexible than current one.

Thanks!

PR #519, opened 2025-06-05
LV2: Automatically detect fallback `showInterface` UI method.

kramlie:
This enables jalv to automatically load the UI correctly for plugins that don't support one of the widget based variants.

### DIFF
--- a/zyngine/zynthian_lv2.py
+++ b/zyngine/zynthian_lv2.py
@@ -520,6 +520,10 @@ def is_plugin_ui(plugin):
                     res = "GtkUI"
                 elif ttl.find("X11UI") > 0 or ttl.find("X11") > 0:
                     res = "X11UI"
+                elif ttl.find("http://lv2plug.in/ns/extensions/ui#showInterface") > 0:
+                    # Fallback UI method when widget based methods are not available.
+                    # See: https://lv2plug.in/ns/extensions/ui#showInterface
+                    res = "UI"
                 else:
                     res = None
                 if res:


jofemodo:
Nice! Thanks @kramlie !

BTW, could you tell me some example to test this?

Thanks

kramlie:
Thanks @jofemodo! To test this, you can just follow the [guide](https://gitlab.com/kramlie/zynthian-yoshimi) I wrote for getting Yoshimi on Zynthian. I'm posting the relevant sub chapter here for convenience:

> Yoshimi
> -------
> 
> Install Yoshimi by **logging into the Zynthian** and executing these commands:
> 
> ```bash
> git clone https://github.com/Yoshimi/yoshimi.git
> cd yoshimi/src
> cmake .
> make -j4
> make install
> ```
> 
> ...
> 
> Execute this command to refresh the available plugins in Zynthian:
> 
> ```bash
> /zynthian/zynthian-ui/zyngine/zynthian_lv2.py
> ```
> 
> We need to enable Yoshimi in the Webconf. Navigate to Engines and then scroll
> down and enable "Yoshimi", which should be near the bottom ("Yoshimi-Multi"
> should not be enabled).

What you need to do:

1. Perform the quoted steps above.
2. Enable the VNC server.
3. Create an instrument chain with the Yoshimi synth.
4. Log into the VNC server, and check that the Yoshimi UI is visible and can be interacted with.

That's it! If the UI works, then the fix is working successfully!

PR #522, opened 2025-06-22
Fix mod-ui autorouting

### LINKED ISSUE #1316: Bug: MOD-UI unnecessary thru audio connections created
**Describe the bug**
When loading the MOD-UI engine, audio connections are created from system capture_1, capture_2 to mon-monitor in_1, in_2 but mod-ui should be managing all of its own connections.

**To Reproduce**
1. Add Special Chain -> MOD-UI -> MIDI CHANNEL 1
2. View the jack connections in Patchage
3. system capture_1, capture_2 is connected to mod-monitor in_1, in_2

**Expected behavior**
system capture_1, capture_2 is not connected to mod-monitor in_1, in_2 unless that connection has been explicitly made in the mod-ui pedalboard screen

**Screenshots**
mod-monitor-connection

**Setup:**
 - Software Version: current oram

**Additional context**
I tried setting audio_thru=False when the engine is "MD" but this had no effect. This bug is a problem for all MOD-UI use because the dry signal is always passed through but there is no visible connection in the mod-ui browser interface. Maybe this needs custom chain setup in the engine start method like for setbfree and aeolus?

Watching Patchage when the engine is started, I can see system capture ports connected to zynmixer input ports, and then mod-monitor is inserted in the middle

steveb:
The logic to avoid connecting to mod-monitor inputs was already there but this must have regressed during the chain rewrite.  This change fixes mod-ui auto routing so that passthrough connections are managed only by the mod-ui graph.

Fixes: https://github.com/zynthian/zynthian-issue-tracking/issues/1316

### DIFF
--- a/zynautoconnect/zynthian_autoconnect.py
+++ b/zynautoconnect/zynthian_autoconnect.py
@@ -915,7 +915,7 @@ def audio_autoconnect():
 
     # Remove mod-ui routes
     for dst in list(required_routes.keys()):
-        if dst.startswith("effect_"):
+        if dst.startswith("mod-monitor:"):
             required_routes.pop(dst)
 
     # Replicate main output to headphones


zynautoconnect/zynthian_autoconnect.py
```diff
@@ -915,7 +915,7 @@ def audio_autoconnect():
 
     # Remove mod-ui routes
     for dst in list(required_routes.keys()):
-        if dst.startswith("effect_"):
```
riban-bw, line 918: We must remove `effect_` nodes from the zynthian graph so that zynthian does not manage these, only mod.

steveb: Ah, I'll modify this to remove both then

jofemodo:
Thanks @steveb

PR #530, opened 2025-07-27
Internet Radio: Add break condition for closed connections. 

g-tezzie:
This PR addresses the high CPU usage of `python3` observed after removing the Internet Radio chain, as reported in https://github.com/zynthian/zynthian-issue-tracking/issues/1437 .

The high CPU usage likely stems from the polling loop within `proc_poll_thread_task()`. It appears `socket.recv()` returns an empty byte array (e.g., `b''`) upon a closed connection, rather than raising a `TimeoutError`. To resolve this, I've added a break condition that exits the loop when `b''` is received, leading to reduced CPU usage in my testing.

Additionally, I've removed the redundant `socket.setblocking(False)` call, as its effect is overridden by the subsequent `socket.settimeout(1)`.


jofemodo:
Seems good to me. @riban-bw , it's your code. please, confirm.

PR #531, opened 2025-08-02
Akai APC Key25 Driver: add support for pattern template

oscaracena:
This PR allows the user to select a pattern as template, which will be automatically used when a new pattern is created. Fixes #zynthian/zynthian-issue-tracking/issues/1441


jofemodo:
I know you know we are thinking of creating a mechanism  similar to this. Not before zynbleton branch  is merged into vangelis, so no problem to implement this at the driver level. We can review it in the future when the mechanism is implemented at lower level.  I hope we get some useful feedback from this early implementation...

oscaracena:
Sure thing! I hope it works in the meantime. :)

PR #532, opened 2025-08-24
Add SooperLooper Mode to AKAI APC KEY25 MK2 driver

pft:
This PR adds an extra mode to @oscaracena 's akai apc key25 mk2 driver to support SooperLooper, without breaking the original nor diverging from it.

Some features:

- unlimited number of loops, of which 4 adjacent ones are be displayed/interacted with at a time.
- loop grouping of up to 5 groups (think verse/chorus etc.)
- quick access to discrete loop functions (rec/overdub/multiply/insert/replace/subst/trig/pan/volume)
- quick access to global and per-loop sync/quantize settings
- quick saving and loading of 40 sooperlooper sessions

The driver is listed as 'AKAI APC Key25 MK2 + SL', next to the original 'AKAI APC Key25 MK2'.

Everything works the same, except that `[Shift+BottomRightPad(note 7)]` enters SooperLooper mode.

Documentation is currently available at
https://github.com/pft/sl-apckey25mk2 (a standalone JS version of just
part of this).

If this gets to be a part of Zynthian proper, I will move the docs to the wiki of course.

## Some technical notes

Since Oscar's driver leaves no pad unused I had to make this a separate driver, sacrificing the function of Shift+the bottom right pad, which likely to be the last one used in some of his (sub) modes (Pattern Editor and Pad Matrix).

For me, this is worth it, but of course not for all. In order to keep up-to-date with any changes in the original driver, and to keep them mostly available, I made a few changes to the original driver.

These changes allow a subclass of the driver to dynamically load more 'handlers' from files in the subdirectory `akai_apc_key25`. A simple dictionary `extraHandlerClasses` lists the notes that are bound to switch to that handler. A handler class should be identical to the module name (filename without '.py') to load, analogous to how ctrldev
drivers themselves are loaded.

If the dictionary is empty no extra handlers are invoked, keeping all the original functionality.


jofemodo:
@oscaracena , could you take a look to this?
I will send you an invitation to zynthian.-ui team so you can make a review if you like.

oscaracena:
Hi @jofemodo. I'll be glad to review the PR. I'll test it on my setup, and check if everything is fine from this part. And I won't say no to that invitation either... 😉

oscaracena:
Hi @pft, and thank you very much for sharing your code! 🙂👍🏻 Nevertheless, I have a couple of questions for you, if you don't mind!

1. Why did you add support for loading dynamic handlers? There is only one defined, and is always loaded in your driver, so it could easily be part of the driver file without much hassle (avoiding the need to dynamic loading).

2. Why do you use a single file for each data class with constants? There is no problem in Python for having more than one class per file, and is also a common practise. It, again, reduces the need to fetch more files from the disk, and keeps the maintaining simpler.

If you don't mind (again), I can make that changes in your code for you. Just let me know.

zyngine/ctrldev/akai_apc_key25/looper_handler.py
```diff
@@ -0,0 +1,2275 @@
...
+                self.shift_down()
+            return
+
+        return
+
+    def get_sessions(self):
+        mypath = self.SL_SESSION_PATH
+        onlyfiles = [
+            f
+            for f in listdir(mypath)
```
oscaracena: This line throws me an error:

```
FileNotFoundError: [Errno 2] No such file or directory: '/zynthian/zynthian-my-data/presets/sooperlooper/'
``` 

You may need to check if that directory exists before trying to load a session.

pft: Oh yeah, should be there, old issue, good point!

pft:
Hi @oscaracena , of course I don't mind. I expect nothing less :). More eyes are always welcome!

1. This seemed the easiest way, but yeah , we could indeed statically load it in the subclass, passing it in a dictionary. The base class would still have to create the instance.
3. Re multiple files for consts: lumping them together in a single file is fine with me.

oscaracena:
Soo... do you want me to make the changes for you? I don't mind either... 😉

pft:
If you don't mind, sure. I am at my paid job right now. Otherwise I'd jump to it!

zyngine/ctrldev/akai_apc_key25/looper_handler.py
pft: BTW, I think @riban-bw once claimed this directory _should_ be auto-created but wasn't. Cannot remember when or where though. And I may be wrong.

oscaracena:
Ok, I've made the suggested changes and it seems to work. Now, I'll try to follow the documentation and use it, just to check that I've not broken it... 😅 The problem is that I've never used SooperLooper before, so it may take me some time.

pft:
Thanks. A different solution than I would have taken, with a pinch of code duplication and super() calls (which I am no hero at), but I understand it if you want no changes to the original driver at all.

I thought you only meant the actual consts, but you incorporated everything in the driver (handler and feedback_leds too).

It may be I want to factor out some again later (with some abstraction to deal with the lower number of colors) once I receive the mk1 controller, but that'll be another story.

At a glance, your changes should work. Good luck looping!

oscaracena:
I'm not very fond of code duplication too, but I propose this as an starting point. I've removed entirely the class `FeedbackLEDs`, which was duplicated. I think `super()` calls are ok for just that, avoid duplicated code, abstraction and also removes coupling. And, in this case, I only see a few lines (in handling the midi event) that are duplicated, but is not too bad.

I would like to also remove the duplicated constants, which are defined in both drivers, but I see that in my original driver are defined in the global space, so are difficult to import. It would be better if the original driver groups them somehow, so you can just import them, is a change we can easily made.

And sorry for troubling you with my changes, I tend to be very aggressive with refactorizations, and I should keep/adhere to other people's way of doing things 😅 Please, feel free to revert anything you feel is better!

oscaracena:
I've tested it following the documentation, and it works great! Very good job @pft, I'll enjoy experimenting with it further! 

Please, add the changes that you wish, and when ready (and if @jofemodo does not say otherwise), we can merge it.

zyngine/ctrldev/akai_apc_key25/looper_handler.py
pft: I added a check, and if `/zynthian/zynthian-my-data/presets/sooperlooper/`  does not exist, it is created it on the fly.

It does not exist if no SL preset had ever been saved from within the zynthian UI.

pft:
@oscaracena I Just tested it. With the change to add the preset directory, which I just committed, it is okay for me to merge this.

Once I get to work on the mk1 (shipped!), I may propose some refactoring to factor out/import the duplicate constants.

oscaracena:
It seems good for me! @jofemodo This PR is ready for merge. Shall I do it? Or do you want to do the honors? 😄

PR #535, opened 2025-09-21
Ableton Push 1 and m-audio keystation pro 88 mk1 Drivers

JBrumby:
Hello everybody,

this is my first try to commit data to an reository. So be calm and peaceful, if I did this wrong.

Now I realize that it would have been better, to do different pull requests.

My changes contain:

1.) zynthian_ctrldev_keystation_pro_88_mk1.py 
This keyboard isn't anymore in production
a comented driver with just 4 knobs to controlling the GUI Interface. Its more a driver example. But it is still iof worth for me, because I use a Zynthian just with touch display without knobs. Parameter cotrol with this is easyer than with touch.

2.) zynthian_ctrldev_base_scale.py 
this is a standarized library to create pad arrays with different scales and modes. It might be useful for more pad driver developers for dumb pad cotnrollers like y Push 1

3.)  zynthian_ctrldev_ableton_push_1_v2.py
This keyboard isn't anymore in production
my main work. It brings my Push 1 to live with Display, LEDs and Pads. Next time I have to learn to use it with Zynthian so I mean, it also could be tested by other Push 1 owners.

I hope I didn't too much wrong with this push request.

Brumby


jofemodo:
Please, remove the not related  files from this pull request:

+ launch.json
+ zynthian.sh
+ zyngine/zynthian_state_manager.py

JBrumby:
ah, I found the "Files changed" section. That is helpful. 
I did it.

jofemodo:
Nice! 

Now you have to remove the "send_midi" stuff from both drivers:

+ For the "pro mk8" it's not needed. Let the zynthian router do the task.
+ For the "ableton push", reimplement the "scales"  functionality as a jack client or mididings, using the new midiproc functionality.

Indeed, the "scale enforcer" is very useful for any controller, so perhaps you could add a generic driver that can be loaded for any controller. We need to think how to "control" the scale used.

Regards,

jofemodo:
Please, could you re-implement the scales functionality for the "ableton push" driver using the latest improvements in ctrldev drivers? You would find a similar example here:

https://github.com/zynthian/zynthian-ui/blob/vangelis/zyngine/ctrldev/zynthian_ctrldev_akai_mpk_moder.py

and the base class for a generic moder (mode enforcer):

https://github.com/zynthian/zynthian-ui/blob/vangelis/zyngine/ctrldev/zynthian_ctrldev_base_moder.py

I would like to include your drivers in the next zynthianOS release.

Thanks!

riban-bw:
This change needs further work as requested by @jofemodo. I would also prefer to see it as seperate PRs for each change. Bundling them all here in a single PR means that an issue with one is blocking the others.

We are in a feature freeze during beta testing of the next stable release (Vangelis). A PR like this that does not touch core functionality but adds compatibility may be considered but we have limited time to test so it only has a chance of inclusion before next stable release if it is very simple for us to validate it works without impacting any other modules. Either way, nothing will happen until the highlighted issues have been resolved.

PR #537, opened 2025-10-02
Fix race condition loading plugins with both controllers and presets.

kramlie:
The race condition can happen when restoring a snapshot, and a plugin both has a preset and controller values. What is supposed to happen is this:

1. Jalv is told to load a preset.

2. Jalv prints the controllers associated with that preset, which is picked up on the process pipe. This causes an update in the UI.

3. New controller values are set based on what was stored in the snapshot, which updates the plugin, and again the UI.

Because the the parsing of the Jalv messages in point 2 happens in a separate thread, it's possible for step 2 and 3 to be reversed. In this case it will cause the UI to go out of sync with the plugin, since the printout from Jalv contains the values from the preset, not the ones that were set most recently.

At that moment it is only a display issue, since the plugin will actually be using the correct values. But the values in the UI are the ones that are saved, so if the snapshot is saved again, the plugin parameters will be wrong the next time the snapshot is loaded.

Fix this by forcing another controller printout from Jalv if we update any controller values while we are still waiting for one.

Fixes:
https://github.com/zynthian/zynthian-issue-tracking/issues/1453

### DIFF
--- a/zyngine/zynthian_engine_jalv.py
+++ b/zyngine/zynthian_engine_jalv.py
@@ -167,6 +167,8 @@ class zynthian_engine_jalv(zynthian_engine):
         self.save_bank = None
         self.save_preset_uri = None
 
+        self.expecting_ctrls = False
+
         if state_manager:
             self.eng_info = self.state_manager.chain_manager.engine_info[eng_code]
         else:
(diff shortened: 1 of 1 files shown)


jofemodo:
This should be already fixed and https://github.com/zynthian/zynthian-issue-tracking/issues/1453 closed, same than other tickets related with this.  Please, re-check again with updated vangelis.

kramlie:
@jofemodo: You're right, it appears to be fixed in vangelis, not in oram.

On vangelis I'm unable to save snapshots though. Completely unrelated of course, but just in case you didn't know. So I'll stick with my patch and oram for now.

Closing this, since it's not needed anymore.

PR #538, opened 2025-10-12
Add plugin specific jalv command line flags

polluxsynth:
Add feature to specify custom jalv command line flags per plugin.

This is used to set the -D flag for MiMi-d, in order to get jalv to set parameters not specified in a loaded preset file to their default values.

This is an example of how to use the new jalv feature implemented in https://github.com/zynthian/jalv/pull/5 .


jofemodo:
OK! I'm testing the jalv mods. As soon jalv is updated with the new mods, i will merge this.

jofemodo:
Merged! 
Vangelis now fully support the -D flag for jalv.

PR #540, opened 2025-10-18
Add driver for Launchkey Mini MK4 37

PatttF:
Driver for Launchkey Mini MK4 37


jofemodo:
Thanks a lot!!
Next time, try to write a better comment for the commit. Think that each commit should have a useful comment and describe the change without having to inspect the commit. For instance: 

`Add driver for Launchkey Mini MK4 37`

would be nice in this case.

PR #542, opened 2025-12-08
Fix forward button mapping for Mackie control and add Zoom R8

jawn:
The button was mapped to cuia_transport_frwd instead of mkc_transport_frwd. Also caused UI log ERROR:zynthian_gui.callable_ui_action: Unknown CUIA 'transport_frwd'.


jawn:
Verified on latest vangelis, works as expected.

PR #543, opened 2025-12-26
Add filter widget and integrate into control UI

ronsum:
This PR introduces a new filter widget for Zynthian. It detects LV2 synth engines that expose cutoff and resonance parameters and displays a dedicated control widget for them.


jofemodo:
I will modify the way you "detect" a filter screen. It must be done by marking the right controls in the TTL using the "designation" property..

PR #545, opened 2026-01-03
Ft leave out ctrldev from zs3 loading

### LINKED ISSUE #1534: Issue: zs3 restores ctrldev driver that was loaded when zs3 was made
# Describe the issue
Using zs3's to conveniently recall parameters and engine presets. Saving them apparently also saves certain states of midi devices, in `midi_capture`. When a MIDI device was connected and had a certain driver loaded at save time, this driver gets recalled when loading a zs3.

This I did not expect and confused, well, something out of me. I expected the current driver to remain loaded and in its current state. Now I have to be very careful to save zs3's in the same state as when I am performing.

Note: there _might_ be a use case for this which I do not yet see, but possibly it is an unwanted result of wanting to restore the midi connections, _in accordance with_ the driver in question.

If there is such a use case, please make it a restore option ('MIDI device drivers': preferably defaulting to off), otherwise, I think the currently loaded driver should be used.

# To Reproduce
## Steps to reproduce the behavior:
1. Load the Fostex Mixtab driver for some MIDI device
2. Save a zs3
3. Load the mackiecontrol driver for the same MIDI device
4. Load the zs3 you just saved
 ## Expected behaviour
The device still set to the mackiecontrol driver
## Actual behaviour
The device gets reset to the Fostex Mixtab driver

# Configuration
## Hardware
```
Raspberry Pi 5 Model B Rev 1.0
Audio: Tascam US-2x2HR
Display: ZynScreen 3.5 (v1)
Wiring: MCP23017_ZynScreen
I2C: MCP23017@0x20
Profile: v4_studio
```
## System
```
Debian GNU/Linux 12 (bookworm)
Build Date: 2024-10-02
Memory: 94% (3821M/4049M)
Internal Storage: 38% (22G/59G ext4)
Temperature: 45.5ºC
Overclock: Medium
```
## MIDI & UI
```
Tuning: 440 Hz
Master Channel: Off
Preload Presets: Off
ZS3 (SubSnapShots): On
Power Save: 10 minutes
Audio Levels on Snapshots: Off
```
## Software
```
zyncoder: oram-2511.3 (e4c91c4) 
zynthian-ui: oram-2511.3 (b4a605b)
zynthian-sys: oram-2511.3 (a596843)
zynthian-data: oram-2511.3 (e4e24b8)
zynthian-webconf: oram-2511.3 (f851967)
```

pft:
Fixes https://github.com/zynthian/zynthian-issue-tracking/issues/1534

This goes to the heart of the Zynthian, and I do not presume to understand everything fully, so please review carefully.

### DIFF
--- a/zyngine/zynthian_state_manager.py
+++ b/zyngine/zynthian_state_manager.py
@@ -1240,6 +1240,12 @@ class zynthian_state_manager:
                 for proc in self.chain_manager.processors.values():
                     proc.set_midi_autolearn(True)
 
+            if load_ctrldev and not load_chains:
+                zs3 = self.sanitize_zs3_from_json(state["zs3"])
+                if not merge:
+                    self.zs3 = zs3
+                self.load_ctrldev(zs3["zs3-0"])
+
             # Save last snapshot info and get snapshot's program number
             self.last_snapshot_count += 1
             if basename(fpath) != "last_state.zss":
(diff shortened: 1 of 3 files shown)


jofemodo:
Hi @pft !
Please, could you explain what are you trying to achieve here?
Also, please, don't send more "changes" to oram branch. Only bugfixes.
All changes, improvements, new features must be done in vangelis now.

pft:
Hi @jofemodo!

Let's say I've made some nice setup with zs3's while having a certain ctrldev driver loaded, say apc key 25 mk2. 

Now, when I change to use APC Key 25 + SL (for using SooperLooper), and I recall a zs3, the driver that I had in use while creating those zs3s suddenly gets loaded instead. 

This PR leaves out the loading of drivers that were running at the time of zs3 creation, except for when loading last state.
It also adds the option to _just_ load those drivers, for the off chance it makes sense. Could be in a variety of settings: quickly load the ctrldev drivers for the stuff you have available at home vs. the stuff in the studio.

Personally I think ctrldev preferences have little to do in a zs3 (except for being a convenient place for the Last State Functionality). But I love to be proven wrong ;). Maybe ctrldev preferences should go somewhere else entirely.

Upcoming PRs will be to vangelis yes. I made this PR before the big zynbleton makeover I think.

jofemodo:
Sorry, but i don't agree.

You can have several drivers for the same device, so you can choose to use some driver in a substate (let's say i want to load the "moder" for my akai MPK mini) but i want to load a different one (or none!) in another substate.

Loaded drivers are part of zynthian state. Without this, state is not complete, so we have to save the drivers state with snapshots and ZS3s.

But ... you could say, "i don't want to **restore** the full state and prefer to omit driver state when restoring".
In such a case, we could improve the ZS3 restoring options (bold click in any ZS3 from the list to access this).

If you agree with this, you could open a feature request, and if you like, implement in this direction.

Kind regards,

PR #547, opened 2026-01-26
Rebuild the optimisation cache after moving a chain

pft:
Fixes  https://github.com/zynthian/zynthian-issue-tracking/issues/1564 for me.


riban-bw (approved):
Looks good to me.

PR #548, opened 2026-01-27
Arturia keylab 61 mk2

trrichard:
This is an early version of the Arturia Keylab 61 MK2 driver. I wanted to get feedback before I moved much farther ahead.

Today it supports basic transport controls, triggering pads and switching between devices. I added all the buttons in case anyone wants to help. 

I'd like to include working faders and some way to switch between the mixer view and pad view in later revisions.

I'd love to hear from the zynthian experts if this is the right direction to go. 

Thank you!

https://www.arturia.com/products/hybrid-synths/keylab-mkii/resources


riban-bw:
We are no longer developing Oram features. All pull requests (except significant bug fixes) should be against Vangelis.

PR #549, opened 2026-02-01
Arturia keylab 61 mk2 driver.

trrichard:
Early support for the arturia keylab 61 mk2.

Faders, encoders and paging have not been implemented yet.


jofemodo:
OK! Merged in Vangelis. You tested in Vangelis, right?
Thanks!

PR #553, opened 2026-03-04
Ft 1586 make arrow keys act consistently in all param editors

pft:
Implements https://github.com/zynthian/zynthian-issue-tracking/issues/1586 as far as I can see.

Enabling arrow keys to edit the values in a param editor (the ephemeral editor at the top of the screen invoked mainly from menus) everywhere on Zynthian.


riban-bw:
It looks like this has been branched from oram. Please branch from vangelis. We are not modifying oram other than critical bug fixes.

PR #554, opened 2026-03-06
fix: map F11 and F12 in keybindings

zieglermarian:
F11 and F12 usually get the keycodes 95 and 96. This minor fix maps these values to 77 and 78 which are used by zynthian-ui for the respective keys. Re-mapping at runtime ensures backwards compatibility.

### DIFF
--- a/zyngui/zynthian_gui_keybinding.py
+++ b/zyngui/zynthian_gui_keybinding.py
@@ -205,6 +205,11 @@ def get_key_action(keycode, modifier):
     logging.debug(
         f"Get keybinding function name for keycode: {keycode}, modifier: {modifier}")
     try:
+        # Hotfix: Map F11 and F12
+        if keycode == 95:
+            keycode = 77
+        if keycode == 96:
+            keycode = 78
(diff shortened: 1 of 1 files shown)


zyngui/zynthian_gui_keybinding.py
```diff
@@ -205,6 +205,11 @@ def get_key_action(keycode, modifier):
     logging.debug(
         f"Get keybinding function name for keycode: {keycode}, modifier: {modifier}")
     try:
+        # Hotfix: Map F11 and F12
+        if keycode == 95:
+            keycode = 77
```
riban-bw, line 210: Keycode 77 is used for num lock and 78 is scroll lock. We use num lock modifier to allow keypads to have 2 layers. This fix does not work.

riban-bw:
This PR fails due to conflict with Num Lock keycode. I think this is the wrong approach. The F11 & F12 keycodes should not be remapped. We should handle them within the keyhandler code.

So I will reject the PR. Take a look at the key handler to see if it can be modified.

PR #555, opened 2026-03-07
Enable arrow up/down for all instances of param editor

pft:
Implements https://github.com/zynthian/zynthian-issue-tracking/issues/1586 as far as I can see.

Enabling arrow keys to edit the values in a param editor (the ephemeral editor at the top of the screen invoked mainly from menus) everywhere on Zynthian.

The other [PR](https://github.com/zynthian/zynthian-ui/pull/553) was not branched of oram, but when creating the PR, github defaults to the default branch, should set vangelis explicitly. Just rebased on latest vangelist for good measure. It's a single commit.


riban-bw:
Thanks for the PR and inspiration. I fixed this in the base class rather than in each child class. Fix is committed to Vangelis.

PR #556, opened 2026-03-11
Fix apckey25 in zynbleton

pft:
Port of ApcKey25 drivers (mk2 and 'gen 1'). Took a long time. Lots of
signatures and screen names changed which had to be adjusted
to. Deepest respect to @oscaracena ; while porting this it struck me
once again what a feat it must have been to write this driver in the first place.

Achieved parity with oram.

## Some functional changes to account for zynbleton changes:

Step sequencer stutter controls now do stutter speed and velocity effects, instead of duration and count.

Scene selection is disabled. User sees a message if invoking those disabled methods.

NOTE: Multiple patterns per sequence pad still supported. Beware the current Zynthian UI has no way to access them.

Blinking tempo button as in v5.

Added a button to invoke the screen 'add_chain' . I find it quite toilsome to get there via the GUI. May remove/change this if v5
implements a good way to access this.

Fixed interfacing the (now two) local alt modes. Added a cuia for pattern editor copy/paste pattern to avoid directly invoking pattern editor methods.

In seqman mode scene selection mode, SHIFT+Up/Down now scroll the view, so we can access phrases beyond the fifth. A message with current scroll value is briefly displayed in the GUI.

## Some fixes to basic zynthian_gui files:

zynthian_gui_pated_base:

Had incorrect signature in start_playback. And `self.wsleds_i_clipboard` can be None.

zynthian_gui_mixer

Improved checking for empty patterns (shell out to libseq method).

Can totally move these into separate PR if so desired of course. They do have their own atomic commits.


oscaracena:
Woow, what a lot of work! Thanks @pft for your kind words, it was very fun indeed! And thanks also for your hard work porting it to match the newest development of Zynthian, much appreciated!!

pft:
Now also implemented the Help function under the bold press of the "alt button".

jofemodo:
Sorry, i can't accept this PR. You have modified critical core files. You shouldn't modify any file out of the ctrldev folder. Any modification to the core API should be discussed with the core developers and managed in a different PR. If they are fixes, please, send a separated PR.

PR #558, opened 2026-03-26
Fix pated alt fn feedback

pft:
wsleds_i_clipboard is None for anything but v5 or v2, in which case it is not enumerable and this code then throws an error.


jofemodo:
Ups! My fault. Thanks for fixing!

PR #560, opened 2026-03-31
fix B3 labels for settings for vibrato and percussion

jlearman:
setbfree: Fix the settings labels for `vibrato selector` and `percussion harmonic` so they match what the controls actually do.

Fixes: #1610


riban-bw:
This will change existing user's snapshots / zs3 but it is the right thing to do so I will merge this PR. (I feel that setBfree has not had the attention it deserves, with several configuration errors like this.) Thanks for fixing this.

jlearman:
I don't think it'll change anything but appearance.  ZSS files have this:

```
            "percussion harmonic": {
              "value": 0
            },
```
I believe we send numbers to setbfree, not names, so the sound will be the same.

PR #562, opened 2026-04-18
Cast clipboard param to int

pft:
Wouldn't work from HID keyboard binding. That is received as a string, therefore no viable index.


riban-bw (approved):
Good catch - thanks!

PR #564, opened 2026-04-29
Fix: Add defensive multitouch capability checks and auto-reconnection…

smespresati:
… for vangelis
Add extended support for usb touchscreen

### DIFF
--- a/zyngui/multitouch.py
+++ b/zyngui/multitouch.py
@@ -234,7 +234,7 @@ class MultiTouch(object):
                 idev = InputDevice(device)
                 idev_caps = idev.capabilities()
                 # Look for the first device supporting multi-touch
-                if idev_caps[ecodes.EV_ABS][ecodes.ABS_Z][0] == ecodes.ABS_MT_SLOT:
+                if ecodes.EV_ABS in idev_caps and ecodes.ABS_MT_SLOT in idev_caps[ecodes.EV_ABS]:
                     self.max_x = idev_caps[ecodes.EV_ABS][ecodes.ABS_X][1].max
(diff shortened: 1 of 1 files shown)


smespresati:
Hi @riban-bw this is the multitouch.py patch to add support to some type of usb touchscreen. In my case, with this patch I didn't need to teak/make any configuration neither configure service or run script at boot time.

riban-bw (commented):
Messages need to be in English.

smespresati:
@riban-bw all messages and comments translated to english

zyngui/multitouch.py
```diff
@@ -270,8 +268,10 @@ def _run(self):
                         self._process_evdev_events()
             except OSError:
                 # Touchscreen driver may have been unloaded so stop thread and enable detection of multitouch (on next xinput touch event)
-                logging.info(f"Multitouch device {self.device_name} disconnected")
-                break
+                logging.warning(f"Multitouch device {self.device_name} disconnected. Retrying in 2 seconds...")
+                import time
+                time.sleep(2)
+                self.open_device()
```
riban-bw, line 274: What was wrong with the previous implementation that put the driver back into auto detection mode after disconnect? This implementation puts a forced 2s delay and an endless loop, even if the device is permanently disconnected.

smespresati: I did it in case disconnecting accidentally or not the device, the procedure will check again for non official touch device when reconnecting. Anyway, I'm not good in programming stuff and, for sure, your magic fingers will optimize the code.

riban-bw:
This change is for DIY builds (non-standard hardware). We are in a feature change freeze beta testing period, so this won't be included during this period. Please remind us a few weeks after the next stable release.

PR #572, opened 2026-05-20
Use internal zynseq method to check emptiness

pft:
The manual way did not work anymore. Cannot remember whether it threw an error or reported incorrectly.

Any way, deferring to dedicated and explicit api is better.

Found this while working on porting apc key25 driver. GUI did not reflect pattern empty/fullness without this.

### DIFF
--- a/zyngui/zynthian_gui_mixer.py
+++ b/zyngui/zynthian_gui_mixer.py
@@ -263,8 +263,8 @@ class zynthian_gui_launcher_pad():
                         n_beats = self.gui_mixer.zynseq.libseq.getBeatsInPattern(pattern)
                         timesig_text = self.get_pattern_length(n_beats, state_phrase["bpb"])
                         try:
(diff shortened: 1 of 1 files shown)


riban-bw (changes requested):
We avoid excessive access of the lib directly because there is a substantial overhead to ctypes lib access. This is why we cache the data in a python dictionary. This is not the right solution. If the `zynseq.state["patns"][str(pattern)]["events"])` cache is invalid, it should be fixed.

jofemodo:
Please, @pft, explain in detail the issue you are trying to fix and give us the steps to reproduce the issue.

riban-bw:
This PR is out of date with merge conflicts and as discussed, it does not fix a known (or fully described) issue in an acceptable way. I am closing this PR. Please revisit the issue you have, raise a ticket and consider the feedback in this PR when implementing a fix. (I suspect the fix should better be within the zynlib.)

PR #573, opened 2026-05-20
Fix apckey25 in zynbleton2

pft:
Port of ApcKey25 drivers (mk2 and 'gen 1') from oram to vangelis.

Second try.  Moved the mildly related changes to core files to separate PRs. So this one only touches files in ctrldev, plus zynthian_gui_config.py, (LAUNCHER_COLOUR) which is intended to be expanded upon.

## Some functional changes to account for zynbleton changes:

Step sequencer stutter controls now do stutter speed and velocity effects, instead of duration and count.

Scene selection is disabled. User sees a message if invoking those disabled methods.

NOTE: Multiple patterns per sequence pad still supported. Beware the current Zynthian UI has no way to access them. I hope a form of arranger will re-emerge.

Blinking tempo button as in v5.

Added a button to invoke the screen 'add_chain'. I find it quite toilsome to get there via the GUI. May remove/change this if v5 implements a good way to access this.

Fixed interfacing the (now two) local alt modes. 

In seqman mode scene selection mode, SHIFT+Up/Down now scroll the view, so we can access phrases beyond the fifth. A message with current scroll value is briefly displayed in the GUI.


jofemodo:
Nice! I don't see any problem now.
I can't test, but others who have the right hardware will do.

Thanks a lot, Niels !

PR #574, opened 2026-05-28
1638/midi cc in zs3

riban-bw:
Adds option to each ZS3 to enable recall of learned MIDI CC binding. Default is disabled. If enabled in all ZS3 it provides the same functionality as before. If disabled in all ZS3 then there is a consistent MIDI learn across all ZS3 which is probably the most desirable workflow and the purpose of this change.


jofemodo:
It seems OK, @riban-bw !
This is a good fix,  mate ;-)

PR #576, opened 2026-06-03
APCKey25: Fix invoking the add chain menu

pft:
zynthian_gui_chain_options no longer has the insert_chain option, calling CUIA instead.


pft:
Superceded by #579

PR #578, opened 2026-06-04
APCKey25: Return True on MIDI CC Events

pft:
There might be a use case for returning False, and letting the Zynthian router take care of the unused CCs (for MIDI learn etc), but this would be undocumented behaviour for these drivers, and letting some arbitrary knobs in arbitrary situations do this is very confusing.

See https://discourse.zynthian.org/t/new-control-device-manager-controller-device-drivers/8593/270


pft:
The reason for this PR is that the driver did not consume all MIDI CC Events, thereby letting them run their normal paths through the router, leading to very unwanted side effects such as changing parameters.

jofemodo:
The main reason to let some CC unconsumed (return False) is to allow this CC to be managed by the MIDI learning system. This is useful and desirable, and used by some ctrldev drivers, like the "Akai MIDI Mix". Same can be said about "Program Change" events and ZS3 suibsystem.

Anyway, the driver decides about this. If you think, for this drivers,  it's better to consume the events with no action at all, avoiding they can be used for MIDI learning, etc., it's perfectly OK.

PR #580, opened 2026-06-29
multitouch: add 90/270 axis-swap rotation via DISPLAY_ROTATION

tunagenes:
# multitouch: add 90°/270° axis-swap rotation via `DISPLAY_ROTATION`

**Base:** `zynthian/zynthian-ui:vangelis`
**Scope:** zynthian-ui only. Additive; no change to existing `None` / `Inverted` behaviour.
**Companion:** a follow-up webconf + zynthian-sys PR (PR 2) adds a Raspberry Pi Touch Display 2 (7″) display definition and exposes the new rotations in the UI. This PR is usable on its own (see "Note on UI exposure" below).

## Problem

`zyngui/multitouch.py` reads the touch evdev device directly and disables the X-layer xinput device, so every X11 / libinput / evdev coordinate transform is bypassed. The driver maps `ABS_MT_POSITION_X → x_root` and `ABS_MT_POSITION_Y → y_root`, with only `invert_x` / `invert_y` available — **there is no axis swap.** A portrait-native DSI panel run in landscape (90° or 270°) therefore has no working touch path: the picture rotates under X, but touch does not follow.

This was hit in practice on a **Raspberry Pi 5 + Raspberry Pi Touch Display 2 (7″)** (portrait-native 720×1280, `ili9881` panel, Goodix `gt911` touch) run in landscape — touch was wrong/dead across half the screen.

## Change

Teach the driver to swap axes, and select the swap/invert flags from the existing `DISPLAY_ROTATION` value.

- `MultiTouch.__init__` gains `swap_xy=False`.
- The X/Y coordinate handling is factored into a pure `MultiTouch._transform_abs(...)` staticmethod. When `swap_xy` is set, a raw panel-X event drives `y_root` and a raw panel-Y event drives `x_root`; the invert flag is chosen by the **destination** axis and the maximum used for inversion is the **source** axis maximum.
- A `MultiTouch.ROTATION_MAP` class constant maps each `DISPLAY_ROTATION` value to `(swap_xy, invert_x, invert_y)`. `zynthian_gui.py` now instantiates the driver from this map instead of the previous hard-coded `Inverted`-only branch.

| `DISPLAY_ROTATION` | rotation | `(swap_xy, invert_x, invert_y)` |
|---|---|---|
| `None`     | 0° (native portrait) | `(False, False, False)` |
| `Right`    | 90° landscape | `(True, False, True)` — **verified on TD2** |
| `Inverted` | 180° | `(False, True, True)` — **unchanged** |
| `Left`     | 270° landscape | `(True, True, False)` — mirror of `Right`, see below |

## Why `DISPLAY_ROTATION` (and not a new variable)

`DISPLAY_ROTATION` is already a touch-only setting:

- In webconf its field is labelled **"Touch Rotation"** (`display_config_handler.py`).
- Its only functional reader is the multitouch instantiation in `zynthian_gui.py` — it has never affected the display.

Display rotation is handled separately (firmware/KMS params and the X11 monitor file generated by zynthian-sys, keyed off `DISPLAY_CONFIG`, not off `DISPLAY_ROTATION`). So adding `Right` / `Left` here completes an existing touch-rotation enum rather than adding new config surface, and it is strictly additive: `zynthian_gui.py` previously special-cased only `Inverted`, with everything else falling through to a plain `MultiTouch(...)`; `None` and `Inverted` map to exactly their previous flag combinations.

I'm happy to switch this to a dedicated variable (e.g. `ZYNTHIAN_TOUCH_SWAP_XY` + invert flags) if maintainers prefer — `multitouch.py` would be untouched and only the small `zynthian_gui.py` selection block would change.

## `Right` is verified; `Left` is not

`Right` `(True, False, True)` was derived from labelled `evtest` corner data on the reference TD2 unit and confirmed working (full width responsive, taps land true). The included test uses that corner data as the oracle.

`Left` (270°) is provided as the exact 180° mirror of `Right` and is checked for that self-consistency in the test, but it has **not** been confirmed on hardware. This is noted in the code (`ROTATION_MAP` comment) and the test. If maintainers would rather not ship an unverified path, `Left` can be dropped to land `Right` only, and added once someone verifies a 270° mounting.

## Tests

New `zyngui/multitouch_rotation_test.py` (runs standalone or under pytest):

```bash
python3 zyngui/multitouch_rotation_test.py
```

Covers: `ROTATION_MAP` keys and the unchanged `None` / `Inverted` flag values (regression lock); `None` = identity; `Inverted` = both axes flipped; `Right` exact mapping against the TD2 corner data plus a per-corner quadrant check ("not mirrored"); and `Left` as the 180° mirror of `Right`. The test stubs `tkinter` / `tkinterweb` / `evdev` / `zynthian_gui_config` so it imports the real class with no GUI, display, or touch hardware present.

## Note on UI exposure (intentional)

This PR does not touch webconf. After it merges, `DISPLAY_ROTATION=Right` / `Left` works but is reachable only by editing `zynthian_envars.sh`; the "Touch Rotation" dropdown still offers only `None` / `Inverted`. Adding those options to the dropdown, plus a TD2 display definition that selects landscape by default, comes in the companion PR (PR 2). The brief window where the capability is env-settable only is intentional.

## Discussion

https://discourse.zynthian.org/t/add-support-for-the-raspberry-pi-touch-display-2-7-inch-screen-to-zynthian-vangelis-feature-request-1704/13332?u=tunagenes

### DIFF
--- a/zyngui/multitouch.py
+++ b/zyngui/multitouch.py
@@ -182,7 +182,47 @@ class MultiTouch(object):
     EVENT_FORMAT = str('llHHi')
     EVENT_SIZE = struct.calcsize(EVENT_FORMAT)
 
-    def __init__(self, state_manager, invert_x_axis=False, invert_y_axis=False):
+    # Map a DISPLAY_ROTATION value to (swap_xy, invert_x, invert_y).
+    #   None     -> 0 deg   (native portrait)
+    #   Right    -> 90 deg  (landscape; verified on RPi Touch Display 2)
+    #   Inverted -> 180 deg (previous behaviour, unchanged)
+    #   Left     -> 270 deg (mathematical mirror of Right; not hw-verified)
+    ROTATION_MAP = {
+        'None':     (False, False, False),
+        'Right':    (True,  False, True),
+        'Inverted': (False, True,  True),
+        'Left':     (True,  True,  False),
+    }
+
+    @staticmethod
+    def _transform_abs(is_x_event, value, swap_xy, invert_x, invert_y, max_x, max_y):
+        """Map a raw ABS_MT_POSITION_{X,Y} event to a (dest_attr, value) pair.
+
+        is_x_event - True for a raw panel-X event, False for a raw panel-Y event.
+        value      - the raw event value.
+        Returns ("x_root" | "y_root", transformed_value).
+
+        When swap_xy is False: raw X -> x_root, raw Y -> y_root (legacy).
+        When swap_xy is True:  raw X -> y_root, raw Y -> x_root (90/270 deg).
+        The invert flag is chosen by the *destination* axis; the maximum used
+        for inversion is the *source* axis maximum (max_x for a raw-X event,
+        max_y for a raw-Y event).
+        """
+        if is_x_event:
+            src_max = max_x
+            if swap_xy:
+                dest, invert = "y_root", invert_y
+            else:
+                dest, invert = "x_root", invert_x
+        else:
+            src_max = max_y
+            if swap_xy:
+                dest, invert = "x_root", invert_x
+            else:
+                dest, invert = "y_root", invert_y
+        return dest, (src_max - value if invert else value)
+
+    def __init__(self, state_manager, invert_x_axis=False, invert_y_axis=False, swap_xy=False):
         """Instantiate the touch driver
 
         Creates an instance of the driver attached to the first multitouch hardware discovered.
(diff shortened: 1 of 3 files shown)


riban-bw:
I would have liked for this to be in the next stable release because it enables more hardware out-of-the-box but this is a DIY (non-standard) hardware configuration so doesn't quite clear the bar for inclusion during this feature freeze beta testing stage - sorry. Please remind us a few weeks after the next stable release (to let the dust settle with bug fixes, etc.)

PR #581, opened 2026-07-01
Add spotify connect chain

pft:
Adds an Audio Generator Chain to serve as Spotify connect. Useful for studying passages (autotuning entire songs haha), making loops, or just listening.

Uses spotifyd under the hood, which needs a premium subscription.

Works only in tandem with the PR https://github.com/zynthian/zynthian-sys/pull/225 on zynthian-sys.

See https://discourse.zynthian.org/t/spotify-connect-chain/13004/6


riban-bw:
This will not be considered before the release of Vangelis due to the current feature freeze. Please remind us of its existance after the next stable release (and few weeks to let the dust settle with bug fixes, etc.)

PR #582, opened 2026-07-04
Fix apckey25 device mode

pft:
This brings back most parts of the device driver @jofemodo fabulously broke yesterday :). 

There were undefined references, a .get() on a list, and incorrect mapping.

Things I did not yet get working:

- Pad for tempo screen/tap tempo does not work yet. And blinking will be interesting too (on active metronome).
- understanding zynswitch in numbers versus led out numbers (they are not the
  same?!). I could not find where these numbers are documented. Some named constants would be awesome.
- understanding arrow keys (should be able to do with zynswitch too?)
- checking/fixing apckey25 (first gen)
- checking/fixing other pads/buttons (dedicated rec/play, zynswitch
  small/large steps on acpkey25 1st gen, more?

NOTE the original driver only lit up rec/play whenever there was something recording/playing. And STOP never. I kinda added that behaviour again, not fully, and wonder whether we should just follow v5 in this respect. Also the pads were just tallies, because there are dedicated rec/play buttons on the device. Need input on that (decision-wise).


jofemodo:
Merged!! I just bought an APC KEY25, so i can test my code ;-)

PR #585, opened 2026-07-15
Implement zynpot_abs in mixer screen

pft:
zynmixer had no zynpot_abs. This addition aids users of pot-based device controllers, such as akai apc key 25 ('mk1').

Ideally, this would have pick-up as the zynthian_gui_controller has, but I found it hard to add an indicator, so now jumps, as most externally fired zctrls to.


jofemodo:
The functionality of knob 4 is invented, but let's try ... it won't harm ;-)

pft:
oh, yes, sorry, forgot to pull that out. But as you say, it won't harm.

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

PR #586, opened 2026-07-15
APCKey25 Send the correct values

pft:
Uses zctrl.midi_control_change in order to translate midi value to zctrl value.

No picking up. this may be better handled in core code. I have some ideas.


jofemodo:
I've implemented a nicer solution.  Please, test & confirm it works in the mk1 MIDI controller.
Implementing value pickup is still pending.

PR #588, opened 2026-07-16
Launchpad pro mk2 vangelis initial fix. Add launcher, device & fader modes.

marvellousmess:
Initial commit. Launcher, device and fader mode working. Note mode not working.

Modes using top buttons as follows:

- Session: Launcher mode. Press again toggle between mixer and launcher views. 
- Note: Not active, needs filtering fixed.
- Device: Provides UI control replicating the 20 button layout of the hardware. 4 buttons at the very bottom replicate encoder presses.
- Fader: Uses launchpad's fader mode to control volume and pan. Use the shift key to switch or volume/pan/sends buttons to go to modes. Send mode not tested and may not work.
- Navigation (arrow) buttons: Work in launcher and device modes. Not working in user mode.

### DIFF
--- a/zyngine/ctrldev/zynthian_ctrldev_launchpad_pro_mk2.py
+++ b/zyngine/ctrldev/zynthian_ctrldev_launchpad_pro_mk2.py
@@ -25,52 +25,790 @@
 # ******************************************************************************
 
 import logging
-from time import sleep
+from time import sleep, monotonic
 
 # Zynthian specific modules
-from zyngine.ctrldev.zynthian_ctrldev_base import zynthian_ctrldev_zynpad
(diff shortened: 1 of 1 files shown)


jofemodo:
Hi @marvellousmess !

I'm merging your changes, but please:

+ Could you consider taking a look to how the APCKey25_mk2 driver implements the device mode? You don't need to re-implement all the logic from the zynthian-UI. Simply re-use it!

+ Regarding the notes mode, how many MIDI ports have the Launchpad Pro mk2? If it have only one, then you would need to implement a "midiproc". If it have 2 ports, you should be using the one port for the notes and another one for the driver modes.

Thanks!

PR #591, opened 2026-07-31
Fix wrong method name and doubling of zynpot + midi learn

pft:
With these changes the top row of knobs in apc key25 is not available for MIDI learn in device mode anymore. Which would be confusing, since this is meant for zynpot. Apc key 25 mk2 driver already did the right thing.

If one wants to use one of the top knobs for midi learn for use with “PAD+CTRL” mode (shift+SEND), the user can start midi learn, switch to PAD+CTRL mode, turn one of those knobs, and it will be bound as usual.

The other change was the addition of zynpot_abs() to gui_base being misnamed as zynpot_cb() (but luckily, the first in order), so ineffective.

I hope to test the changes within the context of the new image soon.


jofemodo:
Thanks @pft !

PR #593, opened 2026-08-12
Optionally cycle zs3

pft:
This allows people to cycle (wrap-around) zs3s with cuia
    zs3_next/zs3_prev using a single pedal.

    Default still stops at the beginning or end.

### DIFF
--- a/zyngine/zynthian_state_manager.py
+++ b/zyngine/zynthian_state_manager.py
@@ -1840,7 +1840,7 @@ class zynthian_state_manager:
 
         return [zs3_id for zs3_id in self.zs3 if zs3_id != "zs3-0"]
 
-    def load_next_zs3(self):
+    def load_next_zs3(self, cycle):
         """Restore the next ZS3, or the first if none is loaded
 
         Returns : True on success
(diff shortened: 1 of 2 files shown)


riban-bw:
This PR adds an optional parameter to the next/prev CUIA commands. This means the CUIA caller must have the capacity to send such a parameter and all CUIA callers must add the parameter for the wrap behaviour. This is flexible but do we envisage users and/or device drivers, etc. wanting to behave differently? If one device driver wraps and another doesn't then the user may be confused by apparently inconsistent behaviour. Would it make more sense to have the wrap option global so that each zynthian behaves consistently to next/prev?

I am open to discussion. I like the flexibility of the proposed solution but worry that in practice it will be more challenging to implement consistent behaviour and may lead to additional support effort.

jofemodo:
Most users using this feature will be mapping CUIA by Master MIDI Channel or OSC. Both options allow to include parameters, so i don't see any problem with this approach. It gives extra flexibility to advanced users.
The CUIA has a default value that should be the preferred option when calling this CUIA from ctrldev drivers, but, ctrldev driver authors have freedom to modify the default. I understand your doubts, @riban-bw, but i don't think this is going to be a big  problem. I would accept the PR.

PR #594, opened 2026-08-29
ZS3 performance view — an ALT mode of the ZS3 screen

KingParamount:
**What this is**

An ALT-mode rendering of the existing ZS3 screen that shows the current cue large enough to read from across a room: a big title, its program-change number, and its position in the running order. Aimed at live performance — my case is a pantomime pit band stepping 30–40 ZS3s from a foot pedal, where the standard list is unreadable at a glance mid-show.

Built as an **ALT mode of `zynthian_gui_zs3`** rather than a new screen, per @jofemodo's suggestion earlier in [thread 13455](https://discourse.zynthian.org/t/is-there-a-performance-mode-for-paging-through-zs3s/13455): a rendering branch on the screen we already have, so it reuses the house topbar (DPM + status area) and the existing `get_alt_mode()` / base-class toggle. No new CUIA, no new screen registration.

**What it adds**

- **`zyngui/zs3_performance.py`** (new) — the view's decisions as plain functions with **no Zynthian imports**: PC label from the ZS3 id (`*/N`, `C/N`, `zs3-N`, bare number), position/total from dict order (the same order stepping walks), ellipsis fitting, and the assembled performance state. Import-free so it's unit-testable without Tkinter or hardware.
- **`zyngui/zynthian_gui_zs3.py`** — the ALT rendering branch: a canvas drawn over the list showing the title (~44 px at 800×480, scaled from the configured font size), the PC label (`CH#…:PRG#…`), and position (`N / M`). It **redraws only on `SS_LOAD_ZS3` / `SS_SAVE_ZS3` and geometry changes** — no polling, and no work in any MIDI or audio path.

**Details**

- A **"NO CUE"** state (nothing loaded yet) that names the first cue in the order.
- Titles **auto-ellipsize to the panel width** (measured with font metrics, never clipped off-screen).
- **Arrow keys step cues while in ALT mode** (down/right → next, up/left → previous, wrapping) — a manual backup if the foot pedal fails or is out of reach. Outside ALT mode the arrows do normal list navigation.
- Fonts and colours come from `zynthian_gui_config`; the view holds no stepping logic of its own — it observes the signals and draws.

**Testing / status**

- Tested on a V5.1 (Pi 5) running `vangelis`, on real hardware.
- The decision module has unit tests in my repo (import-free, no mocks). Happy to move them wherever suits.

**Follow-up (separate PR):** a small change to `save_zs3` to carry a per-cue *note* — as agreed in [the thread](https://discourse.zynthian.org/t/is-there-a-performance-mode-for-paging-through-zs3s/13455) — would add an optional note line to this view. Kept separate so this PR only references fields that already exist.

### DIFF
--- a/zyngui/zs3_performance.py
+++ b/zyngui/zs3_performance.py
@@ -1,0 +1,216 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+# ******************************************************************************
+# ZYNTHIAN PROJECT: Zynthian GUI
+#
+# ZS3 performance view => decision functions
+#
+# Copyright (C) 2026 Charlie Wakely <ccwakely@gmail.com>
+#
+# ******************************************************************************
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of
+# the License, or any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# For a full copy of the GNU General Public License see the LICENSE.txt file.
+#
+# ******************************************************************************
+
+# This module imports nothing. Everything here answers "given this ZS3 dict and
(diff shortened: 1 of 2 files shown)


jofemodo:
Merged! Good work, @KingParamount !

PR #595, opened 2026-09-02
Add an optional note to each ZS3, shown by the performance view

KingParamount:
Follow-on to #594, which added the ZS3 performance view. This adds the one thing that view
was designed around but could not reference until the field existed: an optional free-text
note on a ZS3.

A ZS3 title has to stay short enough to read at a glance, so there is nowhere to put the
things a player actually needs at that moment — which sound the pedal is about to reach, a
tempo, a cue to watch for. I'm running 30–40 ZS3s from a foot controller for a pantomime pit
band, and the note is what turns the performance view from "which cue is this" into something
you can actually play from.

You suggested the ZS3 itself as the right home for it on the forum:

> I think the ZS3 data is the right place for the note. Together with the title and other ZS3
> related info.

So that is where it lives, and it travels with the snapshot without any new state of its own.

## The change

- **`zynthian_state_manager.py`** — `get_zs3_note()` / `set_zs3_note()` beside the title pair,
  and the note is carried across the `save_zs3()` rebuild.
- **`zynthian_gui_zs3_options.py`** — a *Note* entry beside *Rename*, reusing the on-screen
  keyboard.
- **`zs3_performance.py`** — `cue_note()`, and `note` in what `performance_state()` returns.
- **`zynthian_gui_zs3.py`** — the performance view draws it.
- **`zynthian_state_schema.py`** — the new key documented alongside `title`.

Two details worth pointing at:

**`save_zs3()` rebuilds the ZS3 entry from scratch**, so a note has to be carried across an
overwrite the way `title` already is. It follows the existing `restore_midi_learn` pattern
rather than inventing a second one.

**The key is only written when a note is actually set.** A ZS3 with no note has no `note` key,
so existing snapshots are unchanged and nothing has to be migrated.

In the performance view the note shares the line the "no cue" hint uses, since only one of
those two can ever apply — no extra furniture on the screen.

## Tested

On a V5.1 running Vangelis. A note typed into a cue on the touchscreen shows on the
performance face; it survives *Overwrite* on that same ZS3, which is the case this change
exists for, since `save_zs3()` rebuilds the entry; and it survives a full round trip out to
`last_state.zss` and back after a restart. Comparing the snapshot before and after, the only
difference across all eleven ZS3s is the single added `note` key on the one cue, with the ZS3
order unchanged.

One loose end: the *Note* menu entry reuses `zs3_rename.png`, since there is no note icon in
the set. Happy to swap it for a dedicated one if you would like to draw it.

### DIFF
--- a/zyngine/zynthian_state_manager.py
+++ b/zyngine/zynthian_state_manager.py
@@ -1341,6 +1341,32 @@ class zynthian_state_manager:
     def set_zs3_title(self, zs3_id, title):
         self.zs3[zs3_id]["title"] = title
 
+    def get_zs3_note(self, zs3_id=None):
+        """Get ZS3 note
+
+        zs3_id : ZS3 ID (default: Use last loaded zs3)
+        Returns : Note as string, empty if the ZS3 has none
+        """
+
+        try:
+            if zs3_id is None:
+                zs3_id = self.last_zs3_id
+            return self.zs3[zs3_id].get("note", "")
+        except:
+            return ""
+
+    def set_zs3_note(self, zs3_id, note):
+        """Set ZS3 note
+
+        zs3_id : ZS3 ID
+        note : Note as string. An empty note removes the key.
+        """
+
+        if note:
+            self.zs3[zs3_id]["note"] = note
+        else:
+            self.zs3[zs3_id].pop("note", None)
+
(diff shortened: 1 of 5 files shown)


jofemodo:
Merged!!

PR #596, opened 2026-09-02
Give the ZS3 performance view a colour hierarchy

KingParamount:
A follow-up to the ZS3 performance view. It does not depend on #595 (the ZS3 cue note) and the
two can be taken in either order, but #595 is worth reading first: the note is another thing
this hierarchy has to make room for.

Playing from the view on a real stage showed up something the mock-ups never could: everything
on it carries roughly equal visual weight. The cue title is white, the two readouts along the
bottom are yellow and rather loud, and the topbar title is the same white as the cue name. In
a dark pit the eye has to hunt for the one thing that actually matters mid-number, which is
which cue you are on.

This orders the screen by how urgently each part is needed:

- **The cue title takes the yellow**, so it leads. It is the only thing on this screen anyone
  reads at a glance from a distance.
- **The program change and position readouts go green and dimmer.** You need to be able to
  find them when you go looking — "am I at 12 of 38?" — but they should not compete with the
  cue name when you are not.
- **The topbar title dims while the performance face is showing**, and is restored on the way
  out. This uses the per-screen `fg` that `set_title()` already supports, so it touches
  nothing global and no other screen changes.

**The status area is deliberately left alone.** The DPM and its indicators are house furniture
that every screen carries, and a single screen has no business restyling them.

## One thing I would like your opinion on

The readout green is a hardcoded `#008000`, which nothing else on this screen is — every other
colour comes from `zynthian_gui_config`. That is because the palette has exactly one green,
`color_hl` (`#00c000`), and nothing dimmer in that hue, and `color_hl` is bright enough to be
part of the problem being fixed here.

Three ways to resolve it, and I do not want to guess which you would prefer:

1. Use `color_off` (`#5a626d`) instead. Palette-clean, but it is a slate grey rather than a
   green.
2. Add a constant to `zynthian_gui_config`, in keeping with the rest of the palette being
   overridable by `ZYNTHIAN_UI_COLOR_*`.
3. Keep `color_hl` and accept the brighter green.

Say which you would rather have and I will change it — or if you would prefer the colours left
as they are entirely, this PR can simply be closed; the cue note PR does not depend on it.

### DIFF
--- a/zyngui/zynthian_gui_zs3.py
+++ b/zyngui/zynthian_gui_zs3.py
@@ -64,20 +64,26 @@ class zynthian_gui_zs3(zynthian_gui_selector_info):
             0, 0,
             anchor=tkinter.CENTER,
             justify=tkinter.CENTER,
-            fill=zynthian_gui_config.color_tx)
+            fill=zynthian_gui_config.color_ml)
         self.perf_subtitle = self.perf_canvas.create_text(
             0, 0,
             anchor=tkinter.CENTER,
             justify=tkinter.CENTER,
             fill=zynthian_gui_config.color_tx_off)
+        # The one colour here not taken from zynthian_gui_config: the palette
+        # has a single green, color_hl (#00c000), and nothing dimmer in that
+        # hue, and color_hl is bright enough to be part of what this is fixing.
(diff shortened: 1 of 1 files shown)


jofemodo:
I will accept the PR as it's.
Then, i will fix the hardcoded green color, using the color_variant() function.

PR #601, opened 2026-09-11
Route OSC mixer feedback to heartbeat source ports

iibaranov-IG:
A `/mixer/heartbeat` currently registers only `src.hostname`, and the native mixer preconfigures every OSC feedback endpoint with UDP port 1370. Two controllers on the same host therefore collapse into one registration, and changing `--osc-port` cannot redirect feedback.

This tracks clients as `(hostname, source port)` throughout the Oram feedback path:

- the GUI registry uses the heartbeat sender's actual port
- the ctypes wrapper passes that port to the native mixer
- the native endpoint table sends to each registered host-port pair
- timeout removal deletes only the expired endpoint, leaving other ports on the same host active
- duplicate heartbeats for one endpoint remain idempotent

This targets `oram`, where the native OSC feedback path from the report is still present. The later `vangelis` branch removed native feedback and needs a separate migration decision.

BCL evidence: https://github.com/iibaranov-IG/broadcast-control-lab/pull/1

The executable case pins Oram revision `11d4e858367db465a431c701d9035432742a5a7d`, confirms the baseline host-only/fixed-1370 behavior, compiles the patched native sender, and verifies exact OSC datagrams at two independent loopback ports. After removing the first endpoint, the second continues receiving feedback.

Validation:
- Python syntax compilation
- native mixer OSC sender compiled with a JACK test stub
- two-port UDP routing and exact endpoint removal pass under Node 22.20.0
- three repeated loopback runs pass

Hardware verification remains pending. The reporter's three-Zynthian setup can complete it by using distinct `--osc-port` values and confirming fader, mute and DPM feedback at every controller.

Fixes zynthian/zynthian-issue-tracking#1530.

### DIFF
--- a/zyngine/zynthian_engine_audio_mixer.py
+++ b/zyngine/zynthian_engine_audio_mixer.py
@@ -102,6 +102,12 @@ class zynmixer(zynthian_engine):
         self.lib_zynmixer.enableDpm.argtypes = [
             ctypes.c_uint8, ctypes.c_uint8, ctypes.c_uint8]
 
+        self.lib_zynmixer.addOscClient.argtypes = [
+            ctypes.c_char_p, ctypes.c_uint16]
+        self.lib_zynmixer.addOscClient.restype = ctypes.c_int
+        self.lib_zynmixer.removeOscClient.argtypes = [
+            ctypes.c_char_p, ctypes.c_uint16]
+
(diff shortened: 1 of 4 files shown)


jofemodo:
Sorry, we don't accept PRs to Oram. 
Only PRs to Vangelis.

Thanks!

PR #603, opened 2026-09-13
Classic midi CC flag so CC follows note-on events

smiths73v3:
Route MIDI CC according to the CH it was received on, rather than the active channel. This allows CC to follow the note messages.


jofemodo:
Have you tested how this affects global & chain MIDI learning?

PR #605, opened 2026-09-14
Moved pedal and midi indicators up to improve visibility /1749

Octet-nl:
As discussed with @riban in follow up of
https://github.com/zynthian/zynthian-issue-tracking/issues/1749


zyngui/zynthian_gui_mixer.py
```diff
@@ -638,25 +638,26 @@ def __init__(self, parent, canvas, x, width, height, chain, mode):
         self.legend_strip_txt = self.canvas.create_text(self.centre_x, self.gui_mixer.legend_y + self.legend_height / 2, fill=self.gui_mixer.legend_txt_color, text="-", tags=(f"legend_strip_{id}",), font=self.gui_mixer.font)
         self.legend_strip_midi_bg = self.canvas.create_rectangle(x, self.gui_mixer.legend_y + self.legend_height - 2, x + self.width, self.gui_mixer.legend_y + self.legend_height, width=0, fill=self.gui_mixer.legend_bg_color, tags=tags)
 
+        # Moved indicators up to improve visibility
```
riban-bw, line 641: We don't need to comment historical changes to the code.

zyngui/zynthian_gui_mixer.py
```diff
@@ -638,25 +638,26 @@ def __init__(self, parent, canvas, x, width, height, chain, mode):
...
         # MIDI pedal indicators
         self.pedals = []
         for col in range(4):
             self.pedals.append(
                 self.canvas.create_rectangle(
                     int(x + self.width / 5 * col),
-                    self.gui_mixer.legend_y + self.legend_height - 4,
+                    self.gui_mixer.legend_y,
```
riban-bw, line 648: This obscures the record arm indication and the sequence playing progress bar.

Octet-nl: Is this better? I don't know what the `record arm indication` is so I can't test that.
Is that on the same height as the sequence progress bar?

image

Octet-nl: I like this best, just above the legend strip. Stole 4 pixels from the fader height.

Screenshot 2026-09-16 095526

riban-bw:
The record indicator is a small dot that shows in grey when the chain is armed and red when it is recording. You can see this by enabling record arm ('R' button at top of a chain). I just saw that the position of the record indication conflicts with the position of the sequence progress bar.

I like the layout with pedal indications above but I would prefer to avoid eating into the fader and keep all of the status info within the "legend strip" space. Also, I pulled the last change but didn't see the changed behaviour.

Octet-nl:
Git closed this one and rolled back the change
I'm not a great fan of Git....

I changed the positions so the fader stays original length.

image

Created new pr 608.

PR #608, opened 2026-09-16
Moved pedal and midi indicators up to improve visibility /1749

Octet-nl:
Continuation of https://github.com/zynthian/zynthian-ui/pull/605
Getting lost in Git.
I hope this represents the latest version of my repo...


riban-bw:
Sorry - I don't like the visual presentation here. The sequence playback progress bar is now hanging below the top of the frame which feels / looks wrong. It is nice when it is bound to an edge like it was at the top edge before. Also, the progress bar now bleeds over the text in the legend strip. There are a few challenges here that may be difficult to resolve but one of these ideas may work:

- Position the pedal (yellow) and MIDI activity (blue) indicators in a vertical column at one edge (right?) of the legend strip.
- Place the pedal or progress bar above (outside) the legend strip. (Contrary to my previous comment on this). This feels wrong as it may interfere with the view of the fader bottom but it may be okay because the pedal and progress bar are non-permanent / fleeting.

We are trying to fit a lot of info into this area. There may be benefit in defining dedicated status areas so that we can reduce the risk of conflicting status indications.

Octet-nl:
Hi @riban-bw 

`Place the pedal or progress bar above (outside) the legend strip.`

To be honest, I like this most. 
By the way, I didn't phrase it quite right. I didn't reduce the length of the fader; I moved the text up by 4 pixels. I didn't really notice the difference, but it worked fine.

Perhaps a compromise is to move it up by 2 pixels, that gives following picture:

Screenshot 2026-09-17 195230

Your other alternative is also an option:
Screenshot 2026-09-17 200134

Here I couldn't use the full height of the legend strip, if I did, the bottom line would be invisible because of the parallax that started all this ;).

I haven't created a new PR this time; it's better if you decide which option you prefer first. I find the position above the legend the most appealing, but the indicators are less visible when the progress bar is running (I could make then slightly thicker). I don't use the sequencer I keep forgetting the progress bar.

riban-bw:
I quite like the MIDI indicator (blue) being at the bottom of the meters. There is a little bit of correlation between a MIDI event and a change in level. The contrast between the white progress bar and yellow pedal indicators is low which makes this feel abit cluttered due to being indistibuishable. Maybe we can increase the contrast, e.g. darken the pedal indicators. I wonder if we should leave a pixel between each indicator to show they are separate indications rather than a single bar. With the vertical column of indicators, you offset from the bottom to avoid the parallex issue. What would it look like if with the MIDI indicator at the top and occupying the bottom of the meter? This might give more space to the pedal indicators. Also, could we semi-sacrifice the least used pedal at the bottom to the parallex or would that undermine their usefulness?

My concerns are around the visibility of the fader (we don't want a slightly open fader not being indicated) and its control (we must be able to drag the fader all the way closed and reopen - but I think this is coded to allow that anyway). The indications need to be visible and distinct, e.g. with appropriate separation and contrast. Nothing should overlap, including with varied content, e.g. if the legend strip text or symbols are larger, they should still work. It should work on multiple (preferably all) resolutions and must work on the standard official V3, V4, V5 hardware.

I am erring towards a vertical layout, with MIDI indicator above the legend strip (below the meter) and the 4 pedals on the legend strip right hand side below the progress bar with sustain (CC64) at the top with a 1 pixel gap between pedal indicators. The contrast between progress and indicator might want tweaking but it might be okay with a smaller overlap. I appreciate this is not your preferred opton so feel free to offer up both options again with the tweaks made and we can review again.

Octet-nl:
Hi @riban-bw ,

This is the new horizontal option:
Screenshot 2026-09-19 121955

and the vertical one:
Screenshot 2026-09-19 134133

I find the horizontal version more comfortable because my pedals are also side by side. But the visibility of the vertical version is much better.

`semi-sacrifice the least used pedal at the bottom to the parallex`
That is not a problem now (I only have two pedals so I had to change the CC's):
Screenshot 2026-09-19 134041

So it is weighing appearance against functionality, and those two are not far apart.

Since you are the product owner, it is your choice ;)

Edit: This is a 10" 1280x800 HDMI screen. The aspect ratio is a little different, but everything fits (except the font size, but that is out of scope):
afbeelding

riban-bw:
I am going to drag @jofemodo into this now. He has the artisitic eyes that tend to guide style choices. I think the blue MIDI indicator may benefit from a change to improve contrast. (IIRC I used the same colour as the topbar status MIDI indicator.)

This is really good work, thanks @Octet-nl. Let's get the steer from @jofemodo then you can commit and we can reopen this PR. (I don't know why it got closed.)

Octet-nl:
Hi @riban-bw ,

Now the pedal indicators are properly separated the original yellow colour works better. I also brightened the midi colour a bit (here "#19b1f7", can be done from webconf):

image

I can add a new color setting for 'pedal' in webconf if you wish, but I think it is fine this way.

I'll leave it to @jofemodo now.

jofemodo:
I like the latest proposal. 
Why the PR is closed?

riban-bw:
Okay let's go with the horizontal row but align the left pedal indicator with the left of the fader / progress bar and ensure the MIDI indicator is the width of the meters. I think that committing to this PR will allow us to reopen. Ensure you merge latest Vangelis changes

jofemodo:
I just pushed 14aa7d2a that should fix this.

Octet-nl:
Pedal indicators aligned, midi width is not changed. Change committed

riban-bw:
@jofemodo I don't like that implementation. It suffers many of the issues discussed here including floating pedal indications, no gaps between indications, and the MIDI indicator seems to be broken now.

@Octet-nl I see you have reopened the ticket but not pushed any changes so the previous (rejected) implementation remains.

It occured to me that the MIDI indicator makes sense at the bottom of the meters but in launcher mode, the meters move to the top and the MIDI indicator may seem odd, now floating over a launcher.

riban-bw:
Also, there shouldn't be a gap between the progress bar and the pedals. Closing that gap makes both bind to the edge of the fader/legend strip and so nothing is hanging.

Octet-nl:
Hi @riban-bw 

I pushed the the changes: https://github.com/Octet-nl/zynthian-ui/commit/a7735e37d6b3ac9c2987e32d5c26c6abc500d194
But perhaps I do something wrong? I wouldn't know what I have to do now...

Octet-nl:
`the meters move to the top and the MIDI indicator may seem odd`

There is not enough room under the meters at the top. The progress bar and the pedal indicators also remain at the bottom. I don't think that's so strange...

jofemodo:
I just pushed a new proposal. Please, update and test.

Octet-nl:
`new proposal`

The location is fine, but the MIDI indicator now always lights up for the selected chain. It should light up for the chain corresponding to the MIDI channel number being played.

Like this:
image

The pedal's MIDI indicator shows the left chain (channel 1), but the MIDI indicator for the channel being played shows the right chain (channel 2). My keyboard is set to channel 2.

Without this distinction, the indicator is useless (after all, we already have a blue 'm' in the top-right corner of the screen).

riban-bw:
I see the latest iteration has all the indicators at the top which is really odd for a MIDI only chain. These end up floating in the middle of nowhere.

I agree that the change of MIDI indication is flawed. We had an indication of which MIDI channel we being played but now we don't unless the MIDI input is in multitimbral mode. I will have to take a step back and take another look at this to review this functionality. I am getting confused what we are trying to show here.

[Edit] Also - this is a conflation of two features - Flag position and MIDI flag purpose.

riban-bw:
Okay - I have taken another view of the MIDI indicator and I think this current implementation is right (well done @jofemodo). It shows which chain MIDI signals are triggering. That is useful and logical. To know more low-level detail about the MIDI signal, like its MIDI channel you may need to access the MIDI log. There may be advantage to providing a log view within the device itself - but not now. (Feature request!) So let's keep this MIDI indication and concentrate on the flag position.

I don't like that the flags sit in empty space for a MIDI only chain. This looks bad. We can either move them back to the bottom of the fader, immediately adjacent to the progress bar or we could push them right to the top of the channel strip, at the top of the record button, immediately below the top bar.

jofemodo:
I will fix for MIDI only chains ...
Anyway, i will prioritize clarity in synth and audio chains. MIDI only chains are not as widely used.

jofemodo:
This is solved now.

PR #609, opened 2026-09-20
Fix LEDs remaining lighted after UI exit: close race with status thread

tunagenes:
status_thread_task() tests exit_flag only at the top of its loop, so a cycle already in progress finishes and calls wsleds.update() after do_exit() has called wsleds.end(). The repainted frame is latched by the WS281x chain, and on power-off nothing follows to clear it.

Join the status thread before lighting-off the LEDs, and ignore update() once end() has run.

Tested on V5 with touchscreen: reproduced consistently before the patch, 3 of 3 clean power-offs after.


jofemodo:
I 've merged, but the zynthian_gui mod should not be needed.
Indeed, i still can't reproduce the issue. Do you have something weird in your zynthian? Some custom ctrldev driver perhaps?

tunagenes:
No, it is a standard V5 with an RPi4 in it.  The only connections on the back are power, Ethernet, and Analog audio out 1.   Initially the problem surfaced while I was making a cover of Lucy in the Sky with Diamonds and I had 8 chains of instruments loaded.  After a while I cleared all chains and sequences to make the Power off and an faster and the problem stayed.

I do use an Akai key25 with ctrldev driver, the one built into Vangelis, but I've had it disconnected for this testing.

Are you trying to reproduce with an RPi4 based V5?  I've wondered if you're using an RPi5 V5 or V5.1.

